PR #2, opened 2018-08-08
Fix error: 'Could not find `tcp` in `proto`'

zzeroo:
This error occurs if you compile tokio-modbus with the feature 'rtu'.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "tokio-modbus"
 description = "Tokio-based modbus library"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "tokio-modbus"
 description = "Tokio-based modbus library"
-version = "0.2.1"
+version = "0.2.2"
```
flosse: I'd not mix a version inc. with a patch but for now that's ok ;-)

Issue #3, opened 2018-11-30
Windows support


aleksey-r:
**main.rs**
```rust
extern crate futures;
extern crate tokio_core;
extern crate tokio_modbus;
extern crate tokio_serial;

use tokio_core::reactor::Core;
use futures::future::Future;
use tokio_serial::{Serial, SerialPortSettings};
use tokio_modbus::*;

pub fn main() {
    let mut core = Core::new().unwrap();
    let handle = core.handle();
    let port_path = "COM1";
    let server_addr = 0x01;

    let mut settings = SerialPortSettings::default();
    settings.baud_rate = 19200;
    let mut port = Serial::from_path(port_path, &settings, &handle).unwrap();
    port.set_exclusive(false).unwrap();

    let task = Client::connect_rtu(port, server_addr, &handle).and_then(|client| {
        println!("Reading a sensor value");
        client
            .read_holding_registers(0x082B, 2)
            .and_then(move |res| {
                println!("Sensor value is: {:?}", res);
                Ok(())
            })
    });

    core.run(task).unwrap();
}
```

**Cargo.toml**

```rust
[dependencies]  
futures = "*"  
tokio = "*"  
tokio-core = "*"
tokio-serial = "*"
tokio-service = "*"
tokio-modbus = { version = "*", default-features = false, features = ["rtu"] }
```

**Results**
```rust
C:/Users/test/.cargo/bin/cargo.exe run --package modbus_test --bin modbus_test
   Compiling tokio-modbus v0.2.1
error[E0432]: unresolved import `tokio_serial::Serial`
 --> C:\Users\test\.cargo\registry\src\github.com-1ecc6299db9ec823\tokio-modbus-0.2.1\src\service\rtu.rs:3:5
  |
3 | use tokio_serial::Serial;
  |     ^^^^^^^^^^^^^^^^^^^^ no `Serial` in the root

error[E0432]: unresolved imports `tokio_serial::Serial`, `tokio_serial::SerialPortSettings`
 --> C:\Users\test\.cargo\registry\src\github.com-1ecc6299db9ec823\tokio-modbus-0.2.1\src\client.rs:9:20
  |
9 | use tokio_serial::{Serial, SerialPortSettings};
  |                    ^^^^^^  ^^^^^^^^^^^^^^^^^^ no `SerialPortSettings` in the root
  |                    |
  |                    no `Serial` in the root

error[E0433]: failed to resolve. Could not find `tcp` in `proto`
[...]

PR #4, opened 2018-11-30
Restructuring and breaking changes (pre-0.3.0)

uklotzde:
Mostly reorganization of files and module structure behind the scenes, but also some breaking changes:
- Move all public crate exports into a `prelude` module. This revealed many implicit assumptions about public crate exports in the private sub-modules!
- Drop the `ServerType` enumeration and use static typing. You always have to decide for a transport layer when instantiating a service and there is no interaction with the server other than running it.

The README.md contains instructions for migrating existing code from 0.2.x to 0.3.0.

More to follow shortly...


uklotzde:
We may defer and rebase this PR until #3  is fixed.

PR #5, opened 2018-12-01
Internal file and module structure

uklotzde:
- Get rid off the internal `frame::common` module by lifting its contents one level higher into the parent module `frame`. This is consistent with all other modules. Neither functional nor public interface changes.
- All rtu/tcp sub-modules are now contained in separate files
- Some adjustments and extensions of the .gitignore file


uklotzde:
Fixed decoding of incomplete RTU frames (including regression tests)

PR #7, opened 2018-12-03
Follow-up: Fix wrong buffer length check

uklotzde:
This typo does NOT cause any harm, because both the wrong and the correct length bounds are still less than the minimum ADU length.


flosse:
merged with 1bd3549eb47a60d18290c6126009e102c8c52463

PR #26, opened 2019-04-11
use std::convert::TryFrom

flosse:


flosse:
@uklotzde that change is totally boring but in a way I like to celebrate Rust `v1.34.0` ;-)

PR #29, opened 2019-05-16
Add libudev feature to enable/disable tokio-serial/libudev

laumann:
When trying to cross-compile this crate for a Raspberry Pi
(`armv7-unknown-linux-gnueabihf`), the libudev dependency of tokio-serial
failed to build.

I considered two ways of addressing this: Either set up proper
cross-compilation for libudev, or remove it as a dependency. I tried the
latter, and it appears that libudev is not needed as a dependency in
this project.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,17 +22,17 @@ tokio-codec = "0.1"
 tokio-proto = "0.1"
 tokio-service = "0.1"
 
+# Disable default-features to exclude unused dependency on libudev
 tokio-serial = { version = "3.2", optional = true, default-features = false }
 
 [dev-dependencies]
 env_logger = "0.6"
 
 [features]
-default = ["tcp", "rtu", "sync", "libudev"]
+default = ["tcp", "rtu", "sync"]
 rtu = ["tokio-serial"]
 tcp = []
 sync = []
-libudev = ["tokio-serial/libudev"]
 
 [badges]
 travis-ci = { repository = "slowtec/tokio-modbus" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,16 +22,17 @@ tokio-codec = "0.1"
 tokio-proto = "0.1"
 tokio-service = "0.1"
 
-tokio-serial = { version = "3.2", optional = true }
+tokio-serial = { version = "3.2", optional = true, default-features = false }
 
 [dev-dependencies]
 env_logger = "0.6"
 
 [features]
-default = ["tcp", "rtu", "sync"]
+default = ["tcp", "rtu", "sync", "libudev"]
 rtu = ["tokio-serial"]
 tcp = []
 sync = []
+libudev = ["tokio-serial/libudev"]
 
 [badges]
 travis-ci = { repository = "slowtec/tokio-modbus" }


laumann:
Thanks for this project by the way!

flosse:
@laumann Thanks for your PR.
I'd prefer a feature `libudev` that is enabled by default and can be disabled if needed, something like this:

```toml
[features]
default = ["tcp", "rtu", "sync", "libudev"]
rtu = ["tokio-serial"]
tcp = []
sync = []
libudev = ["tokio-serial/libudev"]
```
The dependency of `tokio-serial` might look like this:
```toml
tokio-serial = { version = "3.2", optional = true, default-features = false }
```

I'm not sure if that is working  but that would keep the option to enable/disable the `libudev` features :)

laumann:
> I'd prefer a feature `libudev` that is enabled by default and can be disabled if needed

Sure, I can do that - I just figured that `libudev` wasn't being used, so there'd be no need to keep, but as a feature it lets you keep your options open. The important thing for me right now is the ability to turn it off for `tokio-serial`, and re-exporting it should work.
 
> The dependency of `tokio-serial` might look like this:
> 
> ```toml
> tokio-serial = { version = "3.2", optional = true, default-features = false }
> ```
> 
> I'm not sure if that is working but that would keep the option to enable/disable the `libudev` features :)

That should work - I'm using both `tokio-modbus` and `tokio-serial` as direct dependencies and have specified `tokio-serial` exactly like you show.

laumann:
> > The dependency of `tokio-serial` might look like this:
> > ```toml
> > tokio-serial = { version = "3.2", optional = true, default-features = false }
> > ```
> > 
> > 
> > I'm not sure if that is working but that would keep the option to enable/disable the `libudev` features :)
> 
> That should work - I'm using both `tokio-modbus` and `tokio-serial` as direct dependencies and have specified `tokio-serial` exactly like you show.

Oh, I think I misunderstood what you meant here, but yes, this is what I've done.

flosse:
@laumann Thanks for your contribution. I'll discuss this topic with @uklotzde next week to make a decision.

uklotzde:
The documentation of *mio-serial* states that *libudev* is only needed for port enumeration in [`mio_serial::available_ports`](https://docs.rs/mio-serial/3.2.0/mio_serial/fn.available_ports.html). Since *tokio-modbus* does not invoke or re-export this function we can disable the *libudev* feature as a dependency.

This shouldn't prevent any application from importing *mio-serial* directly with the **libudev* feature enabled.

@laumann Any thoughts? If you confirm those assumptions we would support your initial proposal, i.e. to disable all default features of *tokio-serial*.

laumann:
> The documentation of _mio-serial_ states that _libudev_ is only needed for port enumeration in [`mio_serial::available_ports`](https://docs.rs/mio-serial/3.2.0/mio_serial/fn.available_ports.html). Since _tokio-modbus_ does not invoke or re-export this function we can disable the _libudev_ feature as a dependency.
> 
> This shouldn't prevent any application from importing _mio-serial_ directly with the *_libudev_ feature enabled.
> 
> @laumann Any thoughts? If you confirm those assumptions we would support your initial proposal, i.e. to disable all default features of _tokio-serial_.

@uklotzde That makes sense, I was working off of that assumption when I did my initial proposal. I figured @flosse has some idea for using the `libudev` feature which is why he requested just to re-export it.

In our project, we are importing both `tokio-modbus` and `tokio-serial` like so:

```toml
[dependencies]
tokio-modbus = "0.3.2"
tokio-serial = "3.2.0"
```

before making this PR I tried:

```toml
[dependencies]
tokio-modbus = "0.3.2"
tokio-serial = { version = "3.2.0", default-features = false }
```

but that didn't work.

This tells me that `tokio-modbus` gets compiled with `tokio-serial` w/libudev enabled, and then `tokio-serial` gets compiled with libudev disabled.

I think this confirms your assumption @uklotzde that users that want `libudev` from tokio-serial can just add it as a dependency directly with the feature enabled.

laumann:
I just merged @uklotzde's PR on [omnioiot/tokio-modbus](/omnioiot/tokio-modbus) - for the record, I would just squash the two commits, as I don't see a need to have the re-exported `libudev` in the history anywhere.

uklotzde:
I've squashed the two commits locally and already pushed the result on master. Thanks for discovering and fixing this issue, Thomas! A new release will follow shortly.

uklotzde:
Merged.

laumann:
@uklotzde @flosse Thanks for accepting this change! I'm already using it :smile: The [commit message](https://github.com/slowtec/tokio-modbus/commit/a6bdf915a784f3dacea62895b2db0905bbb23583.patch) got a bit muddled, but the contents work just fine :stuck_out_tongue:

uklotzde:
Ups, I wanted to preserve your commit message verbatim. Maybe a wrong key press during the interactive rebase that accidentally deleted some characters.

PR #39, opened 2020-01-22
Add documentation to the Request type

lovasoa:
The type aliases that are defined in the code are not visible in the documentation, so it's better to have an explicit documentation.

Here is what the current documentation looks like on docs.rs: 
image
https://docs.rs/tokio-modbus/0.3.5/tokio_modbus/prelude/enum.Request.html

The type aliases such as Address, Coil, Quantity, etc. are not visible in the documentation.

### DIFF
--- a/src/frame/mod.rs
+++ b/src/frame/mod.rs
@@ -32,16 +32,59 @@ pub(crate) type Quantity = u16;
 /// A request represents a message from the client (master) to the server (slave).
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum Request {
+
+    /// A request to read multiple coils.
+    /// The first parameter is the address of the first coil to read.
+    /// The second parameter is the number of coils to read.
     ReadCoils(Address, Quantity),
+
+    /// A request to read multiple discrete inputs
+    /// The first parameter is the address of the first discrete input to read.
+    /// The second parameter is the number of discrete inputs to read.
     ReadDiscreteInputs(Address, Quantity),
+
+    /// A request to write a single coil.
+    /// The first parameter is the address of the coil.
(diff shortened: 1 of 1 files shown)


flosse:
> The type aliases such as Address, Coil, Quantity, etc. are not visible in the documentation.

we could also consider to make the type aliases public so we only have to write the documentation once per type.
What do you think?

lovasoa:
Why not even make them real types (structs) instead of type aliases ?

flosse:
> Why not even make them real types (structs) instead of type aliases ?

I don't want to make the API more complex than necessary.
Can you see an added value by using real types beside "only" the documentation?

lovasoa:
> > Why not even make them real types (structs) instead of type aliases ?
> 
> I don't want to make the API more complex than necessary.
> Can you see an added value by using real types beside "only" the documentation?

Just the documentation is already a great benefit in my opinion. The other benefit would be type safety. Currently, it is very easy to write : 

```rust
let address = 1234;
let quantity = 5;
let request = ReadCoils(quantity, address);
```

You code is going to compile just fine, and a code review could very well miss the error.

It would be much harder to miss 

```rust
let address = Quantity(1234);
let quantity = Address(5);
let request = ReadCoils(address, quantity);
```

lovasoa:
But let's maybe discuss that in a separate issue.

flosse:
> But let's maybe discuss that in a separate issue.

:+1:

lovasoa:
Is there something you want me to add or remove before this can be merged?

flosse:
> Is there something you want me to add or remove before this can be merged?

no, it's merged now :)
Thanks for your contribution!

PR #58, opened 2020-04-22
[WIP] Issue 40: read device identification

ptasz3k:
Response and request de- and serialization is done, client api also. I would be grateful for review.

From my current understanding of the codebase the burden of creating server side logic for request processing is on the shoulders of user. In contrast to standard read-write registers functions this one is quite complicated. We have two options, both are not done for the moment:

1. implement all the logic for that function in the library, user task is to initialize the Service with HashMap<int, string> (objId -> value),

2. user handles segmentation and preparation of data, library checks if everything is ok (ascii is ascii, data fits in pdu and so on).

Personally i like option 1 more, but it seems that it doesn't match the 'spirit' of tokio-modbus. What are your thoughts about that?

### DIFF
--- a/src/codec/mod.rs
+++ b/src/codec/mod.rs
@@ -9,7 +9,7 @@ use crate::frame::*;
 
 use byteorder::{BigEndian, ReadBytesExt};
 use bytes::{BufMut, Bytes, BytesMut};
-use std::convert::TryFrom;
+use std::convert::{TryFrom, TryInto};
 use std::io::{self, Cursor, Error, ErrorKind};
 
 impl From<Request> for Bytes {
@@ -64,6 +64,11 @@ impl From<Request> for Bytes {
                     data.put_u16(w);
                 }
             }
+            ReadDeviceIdentification(read_device_id_code, object_id) => {
+                data.put_u8(0x0e);
+                data.put_u8(read_device_id_code);
+                data.put_u8(object_id);
+            }
             Custom(_, custom_data) => {
                 for d in custom_data {
                     data.put_u8(d);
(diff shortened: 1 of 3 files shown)


src/codec/mod.rs
```diff
@@ -115,6 +121,28 @@ impl From<Response> for Bytes {
...
+                next_object_id,
+                device_id_objects,
+            ) => {
+                data.put_u8(0x0e);
+                data.put_u8(read_device_id_code);
+                data.put_u8(conformity_level);
+                data.put_u8(if more_follows { 0xff } else { 0x00 });
+                data.put_u8(next_object_id);
+                /* TODO: do not panic! */
+                data.put_u8(device_id_objects.len().try_into().unwrap());
```
ptasz3k, line 137: Whats the best way to return from here with error without panic?

src/codec/mod.rs
```diff
@@ -115,6 +121,28 @@ impl From<Response> for Bytes {
...
+                for dio in device_id_objects {
+                    data.put_u8(dio.id);
+                    data.put_u8(dio.value.bytes().len().try_into().unwrap());
```
ptasz3k, line 140: Same question as above.

src/codec/mod.rs
uklotzde: Are you still interested in finishing this PR? Then please try to rebase your branch onto #160 which allows this conversion to fail.

ptasz3k:
Hi, I'll try to get back to it till the end of the week. Any thoughts about the question from original pr?

gustavowd:
> stion from original pr?

I can't comment on what the server-side implementation should look like, as I'm not one of the core developers, but I always believe that integrating the library is the best idea.

I'm using your code in a fork on my github for the latest version of tokio-modbus. If you want to verify to use in your pull request, the link is:
[https://github.com/gustavowd/tokio-modbus/tree/device_identification](https://github.com/gustavowd/tokio-modbus/tree/device_identification)

ptasz3k:
Hi, thanks, i'll take a look at it!

src/codec/mod.rs
uklotzde: Ping. See my comment. You could now return an error.

ptasz3k:
@gustavowd Unfortunately, I won't have the time needed to finish this until the end of May. If you want to do it sooner, go ahead!

uklotzde:
What is the plan with this PR?

src/codec/mod.rs
systec-ms: With #292 the option to return an error was removed. Should we re-add the option?

PR #61, opened 2020-05-25
Examples/rtu

zzeroo:
I've shorten the 'rtu-client' example by using the brilliant example from the lib documentation (shorten that example as well a little). The "old" example was a little hard to get from a 'new user' perspective, with all the new async/ .await stuff on top of tokio.

The "old" more complex 'rtu-client' example was restored under a new name.


flosse:
Thanks a lot!

PR #65, opened 2020-09-24
WIP: handle no response for broadcast

cbarber:
Sending a broadcast command should result in a None response instead of panicking. This switches `Result<Response>` to `Result<Option<Response>>` to handle this edge case.

I'm looking for feedback on this approach. Another approach could be splitting `call` into `call_with_response` and `call_without_response` to push the choice of broadcast/non-broadcast farther up to the client


uklotzde:
Fixed by #141.

PR #72, opened 2021-01-03
Upgrade to tokio 1.0.

ivomurrell:
Now that tokio 1.0 is finally out, it seems like a good time to upgrade to the stable version so that other projects can use tokio-modbus after upgrading to tokio 1.0. Currently using the tokio 1.0 update for tokio-serial that is still in PR form so will have to treat this as a draft until that is merged and published. Nevertheless, seems like a fairly straightforward migration!

Closes #71

PR #76, opened 2021-06-10
tokio v1

### LINKED ISSUE #71: Upgrade to tokio 0.3
Hi!

Currently tokio is on version 0.3.6, while this crate only supports the 0.2 versions of tokio.

Using the 0.3 versions of tokio leads to a `thread 'tokio-runtime-worker' panicked at 'there is no reactor running, must be called from the context of Tokio runtime'` runtime error, when using the tcp client to connect to something.

Tarnadas:
Hey,

here is another attempt at updating tokio to v1.
Some remarks:

- I needed to enable the "rt-multi-thread" feature. Otherwise the tcp server would just block and nothing happens.
- I replaced `tokio-serial` with `serial-io`, a custom fork I made based on `mio-serial` and `tokio-serial`. It basically just adds tokio v1 support via a feature flag instead of splitting it into different crates. Also all other dependencies have been updated. I checked [this PR](https://github.com/berkowski/tokio-serial/pull/33) and made sure that I [fixed this](https://github.com/berkowski/tokio-serial/pull/33#discussion_r576928864).
- [serialport-rs](https://gitlab.com/susurrus/serialport-rs) changed its interface to a builder pattern, which is why the API of `serial-io` is slightly different.
- I renamed `Serial` to `AsyncSerial`, because the regular `Serial` in `serial-io` has no tokio support. I'm not sure, if I should change the name.
- I have only tested the TCP async stuff.

Changelog:

- update tokio to v1
- update bytes to v1
- update tokio-util to v0.6
- replace tokio-serial with serial-io
- add sync rtu client example
- replace example cfg flags with cargo toml required features

closes #71 and #72

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,27 +11,27 @@ repository = "https://github.com/slowtec/tokio-modbus"
 edition = "2018"
 
 [dependencies]
-bytes = "0.5"
+bytes = "1"
 byteorder = "1"
 futures = { version = "0.3", optional = true }
 futures-util = { version = "0.3", default-features = false }
 log = "0.4"
 net2 = { version = "0.2", optional = true, default-features = false }
 smallvec = { version = "1", default-features = false }
-# rt-core should be enabled only with "server" feature. Waiting for https://github.com/rust-lang/cargo/issues/5954
-tokio = { version = "0.2", features = ["rt-core"] }
-tokio-util = { version = "0.2", features = ["codec"] }
+# rt should be enabled only with "server" feature. Waiting for https://github.com/rust-lang/cargo/issues/5954
+tokio = { version = "1", features = ["rt", "rt-multi-thread"] }
+tokio-util = { version = "0.6", features = ["codec"] }
 # Disable default-features to exclude unused dependency on libudev
-tokio-serial = { version = "4.3.3", optional = true, default-features = false }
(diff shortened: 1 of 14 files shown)


lovasoa:
Hello !
  We would need this for our work @enexflow. Let us know if there is anything we can do to help this getting merged :) 

@flosse

Tarnadas:
It looks like Windows support for tokio based serial stuff can be worked on once https://github.com/tokio-rs/tokio/pull/3760 is merged. This would be needed for Modbus RTU. I suppose Windows support should not be dropped from tokio-modbus, thus merging is blocked

lovasoa:
Hello !  tokio-rs/tokio#3760  was merged and released.
Is there something else that is blocking ?

ColinFinck:
@Tarnadas I just got the required PR for Windows compatibility into tokio-serial (https://github.com/berkowski/tokio-serial/commit/693e7ee12a368a077fb033e3b434986f4a2d1f4a), with a follow-up PR likely to be merged soon (https://github.com/berkowski/tokio-serial/pull/41)
I expect this to be released as tokio-serial 4.4.0

Could you already restructure your PR around that version of tokio-serial?
This would give us an updated tokio-modbus just in time when the tokio-serial update is on crates.io.

PR #83, opened 2021-07-29
Clean up .gitignore

uklotzde:
Platform-dependent extensions or temporary files caused by custom tools
(like the used editor or IDE) do not belong into the project's .gitignore
file. Instead they must be added to the global .gitignore file which is
placed in the home directory or any other parent directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,18 +1,5 @@
-# Cargo artefacts
+# Cargo artifacts
 Cargo.lock
 
-# Build directory (or a symbolic link to it)
+# Build directory (without a trailing slash to allow using symbolic links)
 /target
-
-# Backup files
-**/*~
-**/*.rs.bk
-
-# macOS
-.DS_Store
-
-# vi files
-*.swp
-
-# VS Code config files
-/.vscode


bootrecords:
While I do not disagree in principle, not everybody sets up a global `.gitignore` like that. I'd assume that it's going to be more tedious to enforce that "clean gitignore policy" within this repository's contributions than it is to just let people add OS- and IDE-related directories to a project's `.gitignore` - from my experience it likely won't do harm, and may even more explicitly prevent contributors from (accidentally) committing such tool config files that can range from irrelevant to irritating for other contributors

flosse:
I can understand both perspectives, but if I had to choose, I would not mix the levels here and keep the `.gitignore` clean.

flosse:
merged in e0378159b7f28cce1f2fe156f47a2535d28280da

uklotzde:
I have noticed that over time everyone tends to dumps their personal preferences into the .gitignore file. It doesn't really matter, but I dislike all this noise that disguises the excludes that really matter for the project.

PR #78, opened 2021-07-22
Yet another Tokio v1 PR

bootrecords:
Based on the PR from @Tarnadas but with `tokio_serial` added back in as a dependency instead of `serial-io` due to its incoming Windows pipes support (from https://github.com/berkowski/tokio-serial/pull/42 and earlier work).
Also homogenized and expanded a bit of documentation for a little bit.

Works on my machine™ (actually semi-field-tested against a serial device from a Linux host and communications appear to work).
Reviews and eventual merge are welcome.

Closes #71 and #76

PR #84, opened 2021-08-02
RTU server

### LINKED ISSUE #73: add rtu to server
As the subject, rtu server is still valuable

bootrecords:
Adds RTU server functionality based on #72

This PR simply replicates that previous proposal that was sidelined by the #78 merge, but updated to compile based on the changes in `tokio-serial`.

Functionality has absolutely not been field-tested by me (beyond the provided example code), and since I have no practical use-case for an RTU server application, I'd ask someone more involved in this field (like @efancier-cn @ivomurrell or @david-mcgillicuddy-moixa) to do the necessary validations.

Closes #73

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,6 +59,11 @@ path = "examples/rtu-client.rs"
 required-features = ["rtu"]
 
 [[example]]
+name = "rtu-server"
+path = "examples/rtu-server.rs"
+required-features = ["rtu", "server"]
+
+[[example]]
 name = "tcp-client-custom-fn"
 path = "examples/tcp-client-custom-fn.rs"
 required-features = ["tcp"]
--- a/examples/rtu-server.rs
+++ b/examples/rtu-server.rs
@@ -1,0 +1,47 @@
+#[tokio::main]
+pub async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    use futures::future;
+    use std::{thread, time::Duration};
+
+    use tokio_modbus::prelude::*;
+    use tokio_modbus::server::{self, Service};
+
+    struct MbServer;
+
+    impl Service for MbServer {
+        type Request = Request;
+        type Response = Response;
+        type Error = std::io::Error;
+        type Future = future::Ready<Result<Self::Response, Self::Error>>;
+
+        fn call(&self, req: Self::Request) -> Self::Future {
+            match req {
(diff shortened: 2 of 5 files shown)


david-mcgillicuddy-moixa:
Thanks for the PR, I'll take a look when I can (hopefully later this week) and test it with our whole stack and hardware.

uklotzde:
Include in v0.5.0?

uklotzde:
Please rebase on master and force push to enable the new CI workflow tasks.

bootrecords:
Hm, looks like the `SerialStream::pair()` function used in one example is only defined on Unix. I'll look into fixes for that later

david-mcgillicuddy-moixa:
> Hm, looks like the `SerialStream::pair()` function used in one example is only defined on Unix. I'll look into fixes for that later

Unfortunately even all the way down into `serialport::COMPort` (which backs `tokio-serial` and `mio-serial`'s `SerialStream` types on windows there's no `pair` function: https://gitlab.com/susurrus/serialport-rs/-/blob/master/src/windows/com.rs#L37

bootrecords:
Well, in that case I see two main options: Either enclosing that example in `#[cfg(target_family = "unix")]` and putting some placeholder for the `not` case, or changing the example to use safe constructors, but then it becomes dependent on connected hardware (not unlike the other RTU examples, actually) whether it would actually run successfully.

david-mcgillicuddy-moixa:
It'll at least compile on windows now, but it seems unlikely to me that this example could possibly work on windows due to "/dev/ttyUSB0" but maybe that's just the best we can do really.

bootrecords:
I was thinking the same thing... that's the trouble with most of the examples (at least on the `rtu` feature) at the moment, though. Somebody may want to spice those examples up with windows/unix `cfg` dependent addresses at some point, perhaps.

david-mcgillicuddy-moixa:
Note that even on a unix platform, namely OSX, I still get the IOCTL errors from `pair` giving me a virtual tty, so bear in mind that isn't a perfect solution either.

flosse (approved):

flosse:
I'm sorry to bug you again but would you mind to rebase this branch again? Otherwise if I'd do it GitHub would not recognize this PR to be merged :-\

flosse:
thx!

PR #85, opened 2021-08-02
Use `async_trait`

### LINKED ISSUE #82: consider using `async_trait`
This is more of an internal structuring thing, but I noticed that in many parts of the code base explicit `Pin<Box<dyn Future<...>>>`s are returned.
While this works, it's a bit tedious to read, so using [async-trait](https://crates.io/crates/async-trait) may be worth considering for the internal traits, such as `Client`, `Reader` and `Writer`.

bootrecords:
First-pass implementation that applies `async_trait` syntax to the `Client`, `Reader` and `Writer` traits in `client/mod.rs` as well as to the `NewContext` trait in `client/util.rs`.

Note that this leaves the explicit `Future` generics from the `Service` trait in `server/service.rs` untouched and also doesn't introduce `async` syntax on functions like `connect[_slave]` in `client/tcp.rs`, `service/tcp.rs` and `service/rtu.rs`. I can still try to work out something there if desired.

Closes #82

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ repository = "https://github.com/slowtec/tokio-modbus"
 edition = "2018"
 
 [dependencies]
+async-trait = "0.1"
 byteorder = "1"
 bytes = "1"
 futures = { version = "0.3", optional = true }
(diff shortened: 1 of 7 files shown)


uklotzde (approved):
Thank you!

PR #89, opened 2021-08-13
Remove `sync` from default features

uklotzde:
Please see the README for a yet unresolved side-effect. I didn't find a workaround for this limitation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## v0.5.0 (Unreleased)
 
+- Removed `sync` from default features
 - Upgraded [tokio](https://tokio.rs/) version from 0.2 to 1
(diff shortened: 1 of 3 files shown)


uklotzde:
Force push after disabling the obsolete Travis and AppVeyor webhooks for the repo.

bootrecords:
It should be able to work around the feature flag issues in doctests as well, [at least if I understand this correctly](https://stackoverflow.com/a/50314936) (I haven't tried it myself, though)

uklotzde:
> It should be able to work around the feature flag issues in doctests as well, [at least if I understand this correctly](https://stackoverflow.com/a/50314936) (I haven't tried it myself, though)

We already use those attributes in lib.rs. But that doesn't prevent the doctests from being executed by `cargo test`.

flosse (approved):
LGTM

flosse:
I also played around with some doctest / features tweaks but without success :(

PR #97, opened 2021-08-26
Modify Server::Service:call signature to include request's unit ID

fdlg:


uklotzde:
@fdlg What is missing to switch the PR from *Draft* to *Ready for review*?

Could you sketch a use case and extend the example accordingly? Currently the additional parameter is ignored.

uklotzde:
Fixed by #141

Issue #57, opened 2020-04-16
Multiple servers on same tokio runtime


madprogrammer:
Hello,

I see this library is being actively developed, and I want to suggest to make it possible to spawn MODBUS TCP slave on the same tokio runtime (reactor) as other servers. I want to make an APP which listens on HTTP, MODBUS TCP and Serial port at the same time, and not to use threads for this purpose. Currently it seems there are not enough abstraction for the TCP server and its startup can't be easily customized. Perhaps you can suggest a way of doing what I need without modifying the `tokio-modbus` library?

flosse:
I think it should be possible but I haven't tried it yet.
We could take your usecase to add it to [examples](/slowtec/tokio-modbus/tree/master/examples) but I won't make it within the next days/weeks.

Would you mind to create draft a PR so I can guide you?

madprogrammer:
So far I got it to work as follows. Not sure how to get rid of `thread::spawn`. I'm just starting to play with Rust, and my original implementation is in Python with Twisted, where it all works well. I just like that Rust produces a single binary which is ready to run on my target ARM-based system, and it's kind of an overhead to deploy all Python stuff there.

```rust
use std::convert::Infallible;
use hyper::{Body, Request, Response, Server};
use hyper::service::{make_service_fn, service_fn};

async fn hello_world(_req: Request<Body>) -> Result<Response<Body>, Infallible> {
    Ok(Response::new("Hello, World".into()))
}

#[tokio::main]
pub async fn main() { 
    use std::net::SocketAddr;
    use std::thread;

    use futures::future;

    use tokio_modbus::prelude::*;
    use tokio_modbus::server::{self, Service};

    struct MbServer;

    impl Service for MbServer {
        type Request = Request;
        type Response = Response;
        type Error = std::io::Error;
        type Future = future::Ready<Result<Self::Response, Self::Error>>;

        fn call(&self, req: Self::Request) -> Self::Future {
            match req {
                Request::ReadCoils(_addr, cnt) => {
                    let mut coils = vec![false; cnt as usize];
                    coils[0] = true;
                    future::ready(Ok(Response::ReadCoils(coils)))
                },
                Request::ReadHoldingRegisters(_addr, cnt) => {
                    let mut registers = vec![0; cnt as usize];
                    registers[1] = 0x77;
                    future::ready(Ok(Response::ReadHoldingRegisters(registers)))
                }
 
                _ => unimplemented!(),
            }
        }
    }


[...]

PR #99, opened 2021-10-30
Update methods on TCP server to be async to allow returning the future

ssnover:
I'm using this crate for a project where we're primarily using the client side APIs, but have recently been working on making "mocked" devices for testing purposes. One difficulty I had was with standing up multiple Modbus servers in a single process since the API calls were internally building their own runtime (this causes a panic if you're already in an async context).

I've revamped the `server::tcp::Server` methods in order for `serve` to be async and for `serve_forever` to call into `serve` for the original use case of a `serve` call that blocks the current thread indefinitely.

One thing I noticed was that the calls related to the `socket2::Socket` seemed to be incompatible in some way with the preconditions expected by `mio`. I stopped using the `configure_tcp` and `listener` calls after discovering that the process was getting blocked on the `accept4` syscall on Linux. It wasn't clear what the intended purpose of those methods were as they seem redundant with setup that `mio` already does or it didn't seem to have an effect when I removed them. I'm happy to revisit though if there's something subtle that's been removed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ tokio-util = { version = "0.6", features = ["codec"] }
 [dev-dependencies]
 env_logger = "0.9"
 futures = "0.3"
-tokio = { version = "1", features = ["net", "macros", "io-util", "rt"] }
+tokio = { version = "1", features = ["net", "macros", "io-util", "rt", "time"] }
 
 [features]
 default = ["tcp", "rtu"]
--- a/examples/tcp-server.rs
+++ b/examples/tcp-server.rs
@@ -1,45 +1,59 @@
-#[tokio::main(flavor = "current_thread")]
-pub async fn main() -> Result<(), Box<dyn std::error::Error>> {
-    use futures::future;
-    use std::{thread, time::Duration};
+use futures::future;
+use std::{net::SocketAddr, time::Duration};
 
-    use tokio_modbus::prelude::*;
-    use tokio_modbus::server::{self, Service};
+use tokio_modbus::prelude::*;
+use tokio_modbus::server::{self, Service};
 
-    struct MbServer;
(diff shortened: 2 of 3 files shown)


ssnover:
This should additionally make it possible to close #57

ssnover:
@flosse Since the TCP server is behind a flag `unstable-tcp-server`, is there a plan for stabilization of the APIs or a roadmap of features? Or are you mainly looking to give the APIs time to get used for people to introduce changes like the one's I've made here?

Either way I'd be happy to help contribute more in this vein.

ssnover:
@flosse Just wanted to bump this to see if you had time to review and give any feedback. Currently pointing Cargo.toml at the branch on my personal repo and would like to at least get it back to trunk of the correct repo!

flosse (approved):

flosse:
@ssnover Thanks for your contribution!
I never really worked with the server side so I'm not an export for that topic but in general it looks good to me :+1: 

@uklotzde Could you also have a look at it if you can find some minutes some when?

flosse:
> @flosse Since the TCP server is behind a flag unstable-tcp-server, is there a plan for stabilization of the APIs or a roadmap of features? Or are you mainly looking to give the APIs time to get used for people to introduce changes like the one's I've made here?

well, as already mentioned we @slowtec never had the server feature in production so it would be irresponsible to call it "stable". But if there are some companies/users who can plausible prove that it is stable, we can rename that feature :wink:

romainreignier:
I have been using this version of the async modbus server for two week as a fake PLC to test my devices. It works great.
It is my first project with tokio and rust async but this version seemed more suited than the upstream version.
Thank you @ssnover for implementing it.

flosse:
@romainreignier thanks for your feedback!

PR #105, opened 2021-12-04
fix an issue when parsing read multiple registers requests in RTU

justinyhuang:
the byte count value resides in the 7th byte of the adu buf, instead of the 5th. this commit fixes it

### DIFF
--- a/src/codec/rtu.rs
+++ b/src/codec/rtu.rs
@@ -433,11 +433,11 @@ mod tests {
         assert_eq!(get_request_pdu_len(&buf).unwrap(), Some(1));
 
         buf[1] = 0x0F;
-        buf[4] = 99;
+        buf[6] = 99;
         assert_eq!(get_request_pdu_len(&buf).unwrap(), Some(105));
 
         buf[1] = 0x10;
-        buf[4] = 99;
+        buf[6] = 99;
         assert_eq!(get_request_pdu_len(&buf).unwrap(), Some(105));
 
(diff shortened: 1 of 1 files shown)


flosse:
@justinyhuang thanks for contributing! Would you mind to adjust the test as well?

uklotzde (changes requested):
I can confirm that according to the [spec,  pp.29-31](https://www.modbus.org/docs/Modbus_Application_Protocol_V1_1b3.pdf) the byte count for function code 0x0F and 0x10 is in the 7th instead of the 5th byte of the ADU. Obviously, no one ever tried to write multiple registers yet. Maybe we got confused by the literal value `4` of the example ;)

Please fix and (if sensible) extend the tests.

uklotzde:
Please amend the commit, because only commits that build without CI failures should be merged.

@flosse We should also backport this fix to v0.4.x after merging.

justinyhuang:
Sure. I will try to update the tests.
Also I tested writing multiple registers with Modbus TCP, which works just fine. So I believe it is just the RTU that is not working correctly.

justinyhuang:
OK I have run `cargo test --features sync` locally and the tests appear passed. 
as a "first-time contributor" I could not kick off CI tests in this repo, thus could you please help approving running the workflows and see if they pass this time?

uklotzde (approved):
Thank you! LGTM

I will squash the commits and adjust the commit message.

PR #106, opened 2021-12-05
[v0.7] Split feature "sync" into "rtu-sync" and "tcp-sync"

uklotzde:
This is required to avoid an unused tokio-serial dependency that is only needed for "rtu-sync" but not for "rtu" (async).

The README and main page of the documentation lists all the public features, including examples on how to use them selectively.


Cargo.toml
```diff
@@ -31,28 +31,39 @@ socket2 = { version = "0.4.7", optional = true, default-features = false }
...
-rtu = ["tokio-serial", "futures-util/sink"]
+default = ["rtu", "tcp"]
+rtu = ["futures-util/sink"]
 tcp = ["tokio/net", "futures-util/sink"]
-sync = ["tokio/rt"]
-server = ["futures", "socket2/all", "tokio/macros", "tokio/rt", "tokio/rt-multi-thread"]
-tcp-server-unstable = ["tcp", "server"]
+rtu-sync = ["rtu", "tokio-serial"]
+tcp-sync = ["tcp"]
+server = ["futures"]
```
flosse, line 50: `dep:futures`

Cargo.toml
```diff
@@ -31,28 +31,39 @@ socket2 = { version = "0.4.7", optional = true, default-features = false }
...
 [features]
-default = ["tcp", "rtu"]
-rtu = ["tokio-serial", "futures-util/sink"]
+default = ["rtu", "tcp"]
+rtu = ["futures-util/sink"]
 tcp = ["tokio/net", "futures-util/sink"]
-sync = ["tokio/rt"]
-server = ["futures", "socket2/all", "tokio/macros", "tokio/rt", "tokio/rt-multi-thread"]
-tcp-server-unstable = ["tcp", "server"]
+rtu-sync = ["rtu", "tokio-serial"]
```
flosse, line 48: I'd use the `dep:` syntax like  `dep:tokio-serial` to make it easier to read and distinguish between features and dependencies.

uklotzde: Will try. In previous versions Cargo complained when using `dep:` without any ambiguities.

uklotzde: Worked.

uklotzde:
The (currently empty) internal feature `"sync"` is needed for #130. Other than before it has not effect when used alone.

uklotzde:
In the long term the synchronous wrapper code should be removed and/or extracted into a separate library: #145

PR #109, opened 2022-01-27
Preallocate vector with exact size when possible

mosolovsa:
https://doc.rust-lang.org/std/vec/struct.Vec.html#capacity-and-reallocation

### DIFF
--- a/src/codec/mod.rs
+++ b/src/codec/mod.rs
@@ -179,7 +179,7 @@ impl TryFrom<Bytes> for Request {
                 if bytes.len() < (6 + byte_count as usize) {
                     return Err(Error::new(ErrorKind::InvalidData, "Invalid byte count"));
                 }
-                let mut data = vec![];
+                let mut data = Vec::with_capacity(quantity as usize);
                 for _ in 0..quantity {
                     data.push(rdr.read_u16::<BigEndian>()?);
                 }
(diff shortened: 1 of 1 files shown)


uklotzde:
Thank you for this easy yet effective improvement. Must have been overlooked.

How about using `usize::from(...)` instead of `as usize` for better type safety? Should work for both `u8` and `u16`.

I'll check the cause for the pre-commit issue soon, hopefully.

uklotzde:
Rebasing on master will hopefully fix the failing pre-commit action.

uklotzde (approved):
Thank you! I'll rebase and merge this PR, the changes are already valid as is.

uklotzde:
Follow-up PR regarding the type casts with a broader scope: #110

mosolovsa:
Thanks for your work and notion about usize::from to prevent possible accidental lossy conversion in the future commits.

I'm not sure should I say this here or open the new issue, I'm curious about is it necessary to hold response as a Vec? There are pretty much use cases when I don't want to have heap allocation on each response, e.g. it would be good if the library user can somehow provide the slice where response should be placed during async machinery. I'm not sure what exactly changes that would require along with API breaking change.

uklotzde:
Sure, there is room for optimization here. Maybe a non-allocating API could be introduced in parallel without affecting existing clients?

`Vec` was probably chosen in the beginning to keep the code simple. And because there was neither a requirement to support `no_std` targets nor real-time environments where Modbus would be unsuitable anyway.

Issue #107, opened 2021-12-15
Cannot run RTU client sync example


sirhcel:
* I wanted to give to give the Modbus RTU client examples a test drive but failed in both cases
    * [`rtu-client-sync.rs`](https://github.com/slowtec/tokio-modbus/blob/5bd83cf660cf79edfe579f91a168e924e73553aa/examples/rtu-client-sync.rs) panics with
        ```
        $ cargo run --features rtu,sync --example rtu-client-sync
           Compiling tokio-modbus v0.5.2 (/.../tokio-modbus)
            Finished dev [unoptimized + debuginfo] target(s) in 0.85s
             Running `target/debug/examples/rtu-client-sync`
        thread 'main' panicked at 'there is no reactor running, must be called from the context of a Tokio 1.x runtime', /.../.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.14.0/src/runtime/context.rs:29:26
        note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
        ```
* Is this examples expected to run out of the box?
* What could I do to actually get it running?
* At a first first glance into [`context.rs:29`](https://github.com/tokio-rs/tokio/blob/f64673580dfc649954eb744eb2734f2f118baa47/tokio/src/runtime/context.rs#L29) it looks like borrowing from `CONTEXT` fails and I bet there is some initialization missing. But this is my first contact with Tokio and I have no idea about the what, how, and why.

sirhcel:
* I meanwhile managed to get [`rtu-client-sync.rs`](https://github.com/slowtec/tokio-modbus/blob/5bd83cf660cf79edfe579f91a168e924e73553aa/examples/rtu-client-sync.rs) running explicitly bringing in a Tokio runtime:
    ```rust
    use tokio::runtime::Runtime;
    fn main() -> Result<(), Box<dyn std::error::Error>> {
        let rt = Runtime::new().unwrap();
        let _guard = rt.enter();

        println!("Hello, world!");
        // ...
    }
    ```
* What's the right way, to get the sync example running?
* I initially erroneously reported the async example hanging (I used the wrong parity configuration) and updated title and description accordingly

manio:
Maybe the example is outdated?
What if you add the line similar as in this example:
https://github.com/slowtec/tokio-modbus/blob/5bd83cf660cf79edfe579f91a168e924e73553aa/examples/tcp-client-custom-fn.rs#L1
[...]

PR #116, opened 2022-06-21
Execute SerialStream::open within an async runtime

sirhcel:
This function internally uses `AsyncFd` and this panics when called without an existing runtime. Seen so in the example `rtu-client-sync.rs` in #107 (with `cfg(unix)`). Runtime::block_on provides the needed runtime now.


uklotzde:
Thanks for contributing. I agree that it doesn't hurt to use `block_on()` unconditionally here.

PR #134, opened 2022-11-11
Fix/tcp delay causing composite tcp packets

thosaa:
I was investigating a problem and noticed that we could trigger the modbus lib to 'collect' multiple Modbus ADUs into a single TCP PDU. 

This is from Wireshark showing the trace between the client (a nodejs script for testing purposes) and the server (baked with tokio-modbus): 
image

Notice that packet 113 has multiple responses, which can be viewed:
image

I am pretty sure that this behavior is causing a problem for one of our customers, who is using a PLC which does support a list of Modbus ADUs in a single TCP PDU. In the other cases our customers PLCs must be waiting for the reponse before making a new request. 

From the [Modbus_Messaging_Implementation_Guide_V1_0b.pdf](http://modbus.org/docs/Modbus_Messaging_Implementation_Guide_V1_0b.pdf) in your README, it seems that this should not be done at all. 
At page 10:
image

This fix is simply to disable Nagle's algorithm by setting nodelay = true. This can not be done the tokio::net::TcpListener directly so I was forced to use from_std and constructing a standard socket.

I was considering making the nodelay a setting, but as it advised not to use it in the document above, I think it should be standard.

This is what it looks like after the fix.
Note that packet 102 / Transaction 5 has 3 requests (Modbus ADUs) in it. The Decoder is perfectly cable of handling this response .
image

### DIFF
--- a/src/server/tcp.rs
+++ b/src/server/tcp.rs
@@ -43,7 +43,6 @@ impl Server {
         S::Instance: Send + Sync + 'static,
     {
         let service = Arc::new(service);
-
         let socket = Socket::new(Domain::IPV4, Type::STREAM, None)?;
         socket.reuse_address()?;
         socket.set_nodelay(true)?;
@@ -121,7 +120,6 @@ where
         }
 
         let request = request.unwrap()?;
-
         let hdr = request.hdr;
         let response = service.call(request.pdu.0).await.map_err(Into::into)?;
 
--- a/src/server/tcp.rs
+++ b/src/server/tcp.rs
@@ -21,6 +21,9 @@ use std::{
 use tokio::net::{TcpListener, TcpStream};
 use tokio_util::codec::Framed;
 
+/// The maximum number of connections queue for acception 
+const CONNECTION_QUEUE_MAX: i32 = 10;
+
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Server {
     socket_addr: SocketAddr,
@@ -46,7 +49,7 @@ impl Server {
         let socket = Socket::new(Domain::IPV4, Type::STREAM, None)?;
         socket.set_nodelay(true)?;
         socket.bind(&self.socket_addr.into())?;
-        socket.listen(1024)?;
+        socket.listen(CONNECTION_QUEUE_MAX)?;
         let listener = TcpListener::from_std(socket.into())?;
 
         loop {
--- a/src/server/tcp.rs
+++ b/src/server/tcp.rs
@@ -43,7 +43,13 @@ impl Server {
         S::Instance: Send + Sync + 'static,
     {
         let service = Arc::new(service);
-        let listener = TcpListener::bind(self.socket_addr).await?;
+
+        let socket = Socket::new(Domain::IPV4, Type::STREAM, None)?;
+        socket.reuse_address()?;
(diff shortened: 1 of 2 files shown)


src/server/tcp.rs
```diff
@@ -43,7 +43,12 @@ impl Server {
         S::Instance: Send + Sync + 'static,
     {
         let service = Arc::new(service);
-        let listener = TcpListener::bind(self.socket_addr).await?;
+        let socket = Socket::new(Domain::IPV4, Type::STREAM, None)?;
+        socket.reuse_address()?;
```
flosse, line 47: What is this for?
Do you mean `socket.set_reuse_address(true)?`

flosse:
Would you mind to fix the CI issues and add a note to the `CHANGELOG.md`?

thosaa:
> Would you mind to fix the CI issues and add a note to the `CHANGELOG.md`?
Fixed fmt.
I added a note and guessed it should be a new release, let me know if it should not.

src/server/tcp.rs
```diff
@@ -43,7 +43,11 @@ impl Server {
         S::Instance: Send + Sync + 'static,
     {
         let service = Arc::new(service);
-        let listener = TcpListener::bind(self.socket_addr).await?;
+        let socket = Socket::new(Domain::IPV4, Type::STREAM, None)?;
```
flosse, line 46: Why is the `protocol` parameter set to `None`? What would happen if we use `Protocol::TCP`?

src/server/tcp.rs
```diff
@@ -43,7 +43,11 @@ impl Server {
...
+        socket.set_nodelay(true)?;
+        socket.bind(&self.socket_addr.into())?;
+        socket.listen(1024)?;
```
flosse, line 49: Could you add some notes why `1024` is used? Ideally point to some documentation somewhere.

flosse:
> > Would you mind to fix the CI issues and add a note to the `CHANGELOG.md`?

> Fixed fmt.
> I added a note and guessed it should be a new release, let me know if it should not.

Thanks!

src/server/tcp.rs
uklotzde: A (private) constant with comments would be appropriate here.

uklotzde:
@flosse We should squash the commits before merging.

src/server/tcp.rs
thosaa: The param to listen is the number of connections allowed queued as I read it from [listen(2)](https://man7.org/linux/man-pages/man2/listen.2.html)
I dont think it's the number of allowed simultanious connections, just "not-yet-accepted" connections. 

I have added a const with comments

uklotzde (changes requested):

src/server/tcp.rs
```diff
@@ -21,6 +21,9 @@ use std::{
 use tokio::net::{TcpListener, TcpStream};
 use tokio_util::codec::Framed;
 
+/// The maximum number of connections queue for acception
```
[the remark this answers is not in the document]
thosaa: Doh! Of cause :-)

uklotzde:
This PR is obsolete now after #159 has been merged. The TCP server accepts a `TcpListener` which could be configured by the application.

PR #141, opened 2023-01-18
Allow server to handle slave address

obayemi:
Hi I've made this small change on the server part of the library to allow it to be more useful

I did need to handle the slave address but it was neither passed to the service nor possible for the service to not respond to a valid modbus request on the bus, wich is a problem on a bus with a lot of devices with their own addresses.

I did not see any PR / issue so I made the fix in a fork that I am currently rolling in production.

It is a bit rough for now but I am interested to know if you would be open to merging it ?

Basically all my changes are to make the `Request` and `Response` types in the services types that need to implement respectively `From<RequestAdu>` and `TryInto<ResponsePdu>`. The change is 100% backward compatible since the original `Request` and `Respone` object implements the required traits, and allow to use any different request / response custom implementation in the service required in the service development that into/from can work with.

Two things I'm not really sure about are:
- having to make `RequestAdu` and `ResponsePdu` public for the interface to work though, but I'm not sure that there actually is a much better way. 
- `TryInto<ResponsePdu>` vs `Option<Into<ResponsePdu>>` for the response type, because I think the `try_into()` is pretty obvious in rust that it is just doing a regular `into()`, but while also taking in account the fact that it might fail, like when no response is supposed to be sent because the message was not destined for that device on the bus, but `Option<Into<ResponsePdu>>` makes it more obvious / limiting that is is actually only designed to allow not sending a response as a success case.

I've not really wrote tests nor yet, but I'm really interested in your feedback.


obayemi:
actually, I realise now that I have missed #97
its intended to do the same thing, but without changing `Service::call` signature, and also allowing to not reply to a request when not relevant.

obayemi:
Added an example in `examples/rtu-server-address.rs`

obayemi:
please @uklotzde @flosse  can you take a look at it ?

Kais-ben-youssef:
+1 Looking forward to this

uklotzde:
Thanks for your contribution!

I am not familiar with the experimental server-side stuff. Maybe someone who actually uses it could spend some time and comment on the changes?

uklotzde:
Hint (for your next PR, not this one): Don't use the main branch in your fork for feature development. Use a feature branch and always pull main from the upstream repo.

obayemi:
> Hint (for your next PR, not this one): Don't use the main branch in your fork for feature development. Use a feature branch and always pull main from the upstream repo.

yeah it was a feature branch on my repo but I merged It when I considered it "done" for my use case (enough for use in my project) I'm doing PRs to review to my changes to myself before sending it here :sweat_smile:

obayemi:
The bulk of the changes is not "really" changes in the server part I just used generics and `From`/`Into` traits to actually make use of thing that were already there. It was mainly just making the `Service` Trait and the servers generic enough to make use of types already present in the project. there are basically only two actual code changes, and a lot of typing stuff to allow it to compile/ 

### Request

The `slaveId` that was already received in `process` in the `RequestPdu` but just not sent in the service call handler that was a non-generic `Request` that didn't contain the id. so I made a new `SlaveRequst` containing the slave and made both `SlaveRequest` and `Request` implement `From<RequestPdu>` instead of having only the `Request` implementing `From<RequestAdu>` wich did not contain the slaveId. and then in process use 
```rs
let response = service.call(request.into()).await.map_err(Into::into)?;
```
instead of
```rs
let response = service.call(request.pdu.0).await.map_err(Into::into)?;
```


### Response

Then I had to make the response optional so the server is able to not repond to request not intended for it (in rtu bus there are lots of devices, on the same bus, the master sends a request with a slaveId, and only the targeted device should reply or it will create conflict on the bus and the message will be corrupted).

So I made the `Response` having to "optionally" need to be `Into<ResponseAdu>` by Using the `TryInto` trait. I think it was easier than having it be either `Option<Into<ResponseAdu>>` or better `Into<Option<Into<ResponseAdu>>>` I think that `TryInto` expresses very elegantly the fact that Response should turn into a `ResponseAdu`, but could also not be, even If the error type is only discarded.

I used `ResponseAdu` and not `ResponsePdu` to allow the service to still send pure `Response` payload if it does not care about the slave, and also to maintain backwards compatibility.

Then the only real change is
```rs
match response.try_into() {
    Ok(pdu) => {
        framed.send(rtu::ResponseAdu { hdr, pdu }).await?;
    }
    Err(_) => {
        debug!("skipping reponse");
    }
}
```
instead of
```rs
framed
    .send(rtu::ResponseAdu {
        hdr,
        pdu: response.into(),
    })
    .await?;
```

uklotzde:
Thank you for the detailed explanations. Sounds reasonable to me.

During reviews I always try to add missing pieces and clarifications to the code. Either by making the code more expressive or adding comments and docs as needed. PR conversations are transient and all the efforts put into them get lost.

Please rebase and fix the failing checks and builds.

src/server/rtu.rs
```diff
@@ -10,6 +10,7 @@ use crate::{
 };
 use futures::{select, Future, FutureExt as _};
 use futures_util::{SinkExt as _, StreamExt as _};
+use log::debug;
```
uklotzde, line 13: I propose to use the `log::` prefix directly in the code instead of importing the module name. Log statements should be easily identifiable. They do not belong to the actual logic.

src/frame/mod.rs
```diff
@@ -239,6 +250,22 @@ impl From<Result<Response, ExceptionResponse>> for ResponsePdu {
...
+where
+    E: Into<ResponsePdu>,
+{
+    type Error = std::io::Error;
+    fn try_from(from: Option<E>) -> Result<ResponsePdu, Self::Error> {
+        match from {
+            Some(e) => Ok(e.into()),
+            None => Err(std::io::Error::new(
+                std::io::ErrorKind::Other,
+                "empty response",
```
uklotzde, line 263: empty response? `None` = no response.

This `TryFrom<Option<..>>` implementation might be convenient, but it seems confusing. Is it really needed?

Where is the context that justifies to reinterpret a neutral `None` as an arbitrary `io::Error`. The generic `TryFrom` trait implementation should not require such context outside of `ResponsePdu`.

obayemi: ok, I'll change the api to use `Option<Into<ResponsePdu>>` instead of the tryfrom then

src/server/rtu.rs
obayemi: on my projects I tend to remove the prefix for logs so I can swich easily from log to tracing, will fix

obayemi:
I think that should fix all the CI errors
and I migrated to using an `OptionalResponse(Option<Response>)` as the `Into` target to make the response be truely optional and not a failure of turning it into an actual response.
also changed logs to use the `log::` prefix

src/frame/mod.rs
```diff
@@ -251,6 +265,18 @@ impl From<Result<Response, ExceptionResponse>> for ResponsePdu {
     }
 }
 
+impl<T> From<Option<T>> for OptionalResponsePdu
+where
+    T: Into<ResponsePdu>,
+{
+    fn from(from: Option<T>) -> OptionalResponsePdu {
```
uklotzde, line 272: - Use `Self` instead of concrete types whenever possible
- Couldn't this expression be simplified? `Self(from.map(Into::into))`

src/frame/mod.rs
```diff
@@ -231,7 +242,10 @@ impl From<RequestPdu> for Request {
 
 /// Represents a message from the server (slave) to the client (master).
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub(crate) struct ResponsePdu(pub(crate) Result<Response, ExceptionResponse>);
+pub struct ResponsePdu(pub(crate) Result<Response, ExceptionResponse>);
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct OptionalResponsePdu(pub(crate) Option<ResponsePdu>);
```
uklotzde, line 248: Why is this type exposed unconditionally if it is only needed for the "server" feature?

uklotzde:
> also changed logs to use the `log::` prefix

Sorry, I didn't notice that it wasn't you who started importing symbols from the `log` module :see_no_evil: Legacy code.

obayemi:
np, good occasion to fix the legacy code :)

src/frame/mod.rs
obayemi: good point, I have to say I'm not yet completely accustomed to just how flexible the feature system is in rust and it didn't even occur me that it was possible to conditionally publish it, I'll fix it later

obayemi: hmmm, I actually ended up making the whole declaration only with the server feature :shrug:

obayemi:
and now that I finally figured out that the test suite was in pre-commit all along, this should fix the last tests

src/server/rtu.rs
```diff
@@ -90,13 +94,20 @@ where
...
+            .await
+            .map_err(Into::into)?
+            .into();
+
+        match response.0 {
+            Some(pdu) => {
+                framed.send(rtu::ResponseAdu { hdr, pdu }).await?;
+            }
+            None => {
+                log::debug!("skipping response");
```
uklotzde, line 108: This log message could be rewritten to better reflect the context and outside view instead of reflecting implementation details. How about `"No response for request {hdr:?}"`

Personal preference: I tend to start log messages with a capital letter, like a Git commit message.

src/server/tcp.rs
```diff
@@ -116,14 +117,20 @@ where
...
+            .await
+            .map_err(Into::into)?
+            .into();
+
+        match response.0 {
+            Some(pdu) => {
+                framed.send(tcp::ResponseAdu { hdr, pdu }).await?;
+            }
+            None => {
+                log::debug!("skipping response");
```
uklotzde, line 131: See above

uklotzde (changes requested):

src/prelude.rs
```diff
@@ -33,9 +33,11 @@ pub mod tcp {
 pub use crate::server;
 
 ///////////////////////////////////////////////////////////////////
+#[cfg(feature = "server")]
```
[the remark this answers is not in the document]
obayemi: fuck... that's rustfmd doing it for who knows why... I even disabled it in my editor to update it but it seems it still found a way.

For the exports, actually, since in the interface, `OptionalResponsePdu` is only used as an `Into<>` bound, it is not required to export its inner ResponsePdu, since Using a `Response` or `Option<Response>` in the actual Services works.

And yes, I'm removing it from prelude. I think I had a clippy warning at some point about it not requiring to be pub because it was not exported that made me put it in prelude but I cant see it anymore... I'm completely removing it from prelude :+1:

src/server/rtu.rs
obayemi: yes, logging the hdr is a good idea, I wanted to log the error from TryInto but I gave up and forgot :sweat_smile:

src/prelude.rs
```diff
@@ -35,7 +35,7 @@ pub use crate::server;
 ///////////////////////////////////////////////////////////////////
 /// Structs
 ///////////////////////////////////////////////////////////////////
-pub use crate::frame::{Request, Response};
+pub use crate::frame::{Request, Response, SlaveRequest};
```
uklotzde, line 38: My remarks also apply to `SlaveRequest` that should neither be part of the `prelude` nor visible without the "server" feature.

obayemi: if I don't put it in prelude it won't compile because the frame mod is not publicly exported in `lib.rs`
I don't know what would be the best way to handle this for this project. Do you have any recommendation ?

uklotzde: Good point. Then keep it in the prelude and hide it behind the "server" feature.

src/frame/mod.rs
```diff
@@ -105,6 +107,15 @@ pub enum Request {
     Disconnect,
 }
 
+/// A Modbus request with slave included
+#[derive(Debug, Clone, PartialEq, Eq)]
```
uklotzde, line 111: The feature flag is still missing. Please also note that you need to reference `SlaveId` with the full path and remove the `use` statement to avoid warnings.

uklotzde: Even if the module is private, the feature flag is for documentation and separation of concerns.

obayemi: Sorry for being slow, I'm not very used to actually making open source contributions :sweat:  (especially in rust)

uklotzde (approved):
Thank you! I'll add a changelog entry.

obayemi:
already on crates !
awesome thx :)

PR #142, opened 2023-01-26
Add MaskWriteRegister

JosuGZ:
This is an implementation of the masked write register function code (0x16).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@
 
 # Changelog
 
+## v0.5.5 (2023-01-30)
+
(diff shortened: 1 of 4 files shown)


uklotzde:
Thank you! Would you mind adding a changelog entry? I guess this could be released as v0.5.5.

LGTM. Anyone else?

uklotzde (approved):

JosuGZ:
> Thank you! Would you mind adding a changelog entry?

Done!

uklotzde:
Thanks for the hint. This is indeed a breaking change, so we better release it as v0.6.0. Version numbers are not a scarce resource ;)

PR #151, opened 2023-02-24
First TLS implementation

gustavowd:
First tokio modbus implementation to the new specification Modbus TCP Security, based on TLS.


uklotzde:
- Certificates (especially private keys) do not belong into the repo.
- Heavy duplication of code for implementing a cross-cutting concern doesn't seem to be right.

If it works for your use case you may use a private fork until someone comes up with generic and extensible solution that is suitable for everyone.

I recommend to close this PR. Otherwise someone else please step in to review it.

gustavowd:
I believe the project should have an implementation of the new Modbus Secure standard, because since 2019 new products with Modbus TCP are not recommended due the various security issues. Certificates were added for test purposes, but I can remove them. The implementations of the MODBUS TCP and RTU servers already have a lot of code duplication, so I followed the project pattern. If the project maintainers believe that my code is not generic enough, I recommend that there is soon an implementation of this standard which was disclosed in 2018.

uklotzde:
TCP and RTU are different transport layers that require different frames and codecs.

TLS only affects the way how TCP connections are established, but otherwise both frame and codec are unaffected. Please eloborate if this is not the case. I don't see any difference in the code you submitted.

gustavowd:
> 

I got your point. I reduced code duplication considerably by removing the tls codec and frame implementation files, which really could be the same as tcp.

gustavowd:
I removed the Certificates as well.

gustavowd:
I beleive that now the code is 100% compatible with Modbus Security and has a minimal of code duplication.

uklotzde:
- The server-side implementation could now be provided as an example after #159 has been merged.
- The client-side API modifications required for supporting TLS connections are an open issue, see #161.

Hint: Loading certificates from the file system is very limiting even without hard-coded paths. What works for your particular use case might not work for all other use cases.

PR #156, opened 2023-02-28
sync: Fix panic when using client-side timeouts

### LINKED ISSUE #155: crash when using sync::rtu with timeout
This code doesn't set a timeout, even if SerialPortBuilder timeout is set:

```
    let builder = tokio_serial::new(config.serial_port(), config.baud_rate())
        .timeout(Duration::from_secs(3))
        .parity(Parity::None)
        .data_bits(DataBits::Eight)
        .stop_bits(stop_bits);
    let mut context = sync::rtu::connect(&builder)?;
```

We can however set the timeout manually in the context:
```
    context.set_timeout(Duration::from_secs(3));
```

The problem is that it crashes later when calling block_on_with_timeout with:
```panicked at 'A Tokio 1.x context was found, but timers are disabled. Call `enable_time` on the runtime builder to enable timers.'```

The most probable fix would be to add enable_time when creating the runtime in client/sync/rtu.rs, connect_slave_with_timeout:
```
    let runtime = tokio::runtime::Builder::new_current_thread()
        .enable_io()
        .enable_time() // add this here
        .build()?;
```

uklotzde:
- v0.7.1
- Fixes #155 
- tcp-server-unstable: Upgrade socket2 dependency (marked as _unstable_, so this possibly breaking change doesn't matter)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 ## v0.7.1 (2023-02-28)
 
 - Fix (sync): Panic when using client-side timeouts [#155](https://github.com/slowtec/tokio-modbus/issues/155)
+- tcp-server-unstable: Upgrade socket2 dependency
 
 ## v0.7.0 (2023-02-14)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,11 @@
 
 # Changelog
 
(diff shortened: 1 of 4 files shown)


flosse (approved):

uklotzde:
I'll tag and publish to crates.io.

PR #158, opened 2023-03-03
Now the unit id can be accessed on the server when responding a request

gustavowd:
Now the unit id can be accessed on the server when responding a request by the call function. This is important for implementing gateways as one tcp or tls server can be a gateway for multiple modbus rtu servers.


uklotzde:
This has already been implemented in v0.6.1 by introducing `SlaveRequest`. Please re-open if this is not the case.

gustavowd:
Thanks. Yes, that's the case. SlaveRequest can be used to get the unit id.

PR #159, opened 2023-03-04
Revise server features

### LINKED ISSUE #115: API Design on unstable tcp server: purpose of NewService trait
Hello,

I'm running into some API ergonomic issues when trying to use the TCP server module part of this crate. I have attempted to extend the example `tcp-server.rs` with a new method: 

```rust
impl MbServer {
    async fn run(self, server: server::tcp::Server) -> std::io::Result<()> {
        server.serve(self).await
    }
}
```

The first issue I run into is this compiler error:
```
error[E0277]: expected a `Fn<()>` closure, found `MbServer`
  --> examples/tcp-server.rs:31:22
   |
31 |         server.serve(self).await
   |                ----- ^^^^ expected an `Fn<()>` closure, found `MbServer`
   |                |
   |                required by a bound introduced by this call
   |
   = help: the trait `Fn<()>` is not implemented for `MbServer`
   = note: wrap the `MbServer` in a closure with no arguments: `|| { /* code */ }`
   = note: required because of the requirements on the impl of `NewService` for `MbServer`
```
Fair enough, the architecture is built around the idea of a "callable" service. So I refactored to:
```rust
server.serve(|| Ok(self)).await
```
But then it gets very much in the way:
```
error[E0525]: expected a closure that implements the `Fn` trait, but this closure only implements `FnOnce`
  --> examples/tcp-server.rs:31:22
   |
31 |         server.serve(|| Ok(self)).await
   |                ----- ^^^^^^----^
   |                |     |     |
   |                |     |     closure is `FnOnce` because it moves the variable `self` out of its environment
   |                |     this closure implements `FnOnce`, not `Fn`
   |                the requirement to implement `Fn` derives from here
```
As I'm moving the struct implementing Service here, this restriction on the API seems unnecessary.

While digging to figure out why the API requires this, I stumbled on the `NewService` trait and it's not clear to me what it's role was. I adjusted the functionality of `serve` et al to bound `S` as `Service` instead of `NewService` and my original implementation compiled fine and functioned as I expected.

It seems a lot of complexity and extra restrictions are added as a result of this `NewService` trait with no obvious benefit. Perhaps I'm misunderstanding the use case, but as the API is unstable, would it perhaps make sense to remove it or add more functions for addressing the extra complexity of the other use case only when necessary?

### LINKED ISSUE #133: MODBUS/TCP Security support (TLS)
Hi. Congrats for the project.
Is it possible to add support for the new specification, the Modbus TCP Security, based on TLS?

Some of new protocols based on modbus are migrating for this modbus version, as the sunspec modbus protocol. The specification is available at:
<https://modbus.org/docs/MB-TCP-Security-v21_2018-07-24.pdf>

I beleive it can be implemented based on tokio-rustls.

Thanks in advance.
Gustavo Denardin

uklotzde:
The current design is both too opinionated and inflexible.

- Stabilizes "tcp-server-unstable" as "tcp-server"
- Obsoletes #134
- Fixes #133
- Obsoletes #151
- Fixes #115

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -39,7 +39,7 @@ jobs:
 
     steps:
       - name: Install Rust toolchain
-        uses: dtolnay/rust-toolchain@master
+        uses: dtolnay/rust-toolchain@v1
         with:
           toolchain: stable
           targets: ${{ matrix.target }}
--- a/.github/workflows/pre-commit.yaml
+++ b/.github/workflows/pre-commit.yaml
@@ -29,7 +29,7 @@ jobs:
           python-version: "3.x"
 
       - name: Install Rust toolchain
-        uses: dtolnay/rust-toolchain@master
+        uses: dtolnay/rust-toolchain@v1
         with:
           toolchain: stable
           components: clippy, rustfmt
--- a/.github/workflows/security-audit.yaml
+++ b/.github/workflows/security-audit.yaml
@@ -20,7 +20,7 @@ jobs:
   security_audit:
     runs-on: ubuntu-latest
     steps:
-      - uses: dtolnay/rust-toolchain@master
+      - uses: dtolnay/rust-toolchain@v1
         with:
           toolchain: stable
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,7 +49,7 @@ repos:
       - id: check-github-actions
       - id: check-github-workflows
   - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: v3.0.0-alpha.5
+    rev: v3.0.0-alpha.6
     hooks:
       - id: prettier
         types_or:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,7 +49,7 @@ repos:
       - id: check-github-actions
       - id: check-github-workflows
   - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: v3.0.0-alpha.4
+    rev: v3.0.0-alpha.5
     hooks:
       - id: prettier
         types_or:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,16 @@
 
 # Changelog
 
+## v0.8.0 (Unreleased)
+
+- TCP Server: Replace `NewService` trait with closures and never panic
+- TCP Server: Stabilize "tcp-server" feature
+- RTU Server: Remove `NewService` trait and pass instance directly
+
 ## v0.7.1 (2023-02-28)
 
 - Fix (sync): Panic when using client-side timeouts [#155](https://github.com/slowtec/tokio-modbus/issues/155)
-- tcp-server-unstable: Upgrade socket2 dependency
+- tcp-server: Upgrade socket2 dependency
 
 ## v0.7.0 (2023-02-14)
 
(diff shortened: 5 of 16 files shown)


flosse (approved):
I did not test it but it looks good to me!

gustavowd:
Sorry, but I don't get it how this PR can fixes [MODBUS/TCP Security support #133](https://github.com/slowtec/tokio-modbus/issues/133) issue.

uklotzde:
> Sorry, but I don't get it how this PR can fixes [MODBUS/TCP Security support #133](https://github.com/slowtec/tokio-modbus/issues/133) issue.

You could configure `TcpListener` in the outer context. That does not belong to this library.

gustavowd:
> > Sorry, but I don't get it how this PR can fixes [MODBUS/TCP Security support #133](https://github.com/slowtec/tokio-modbus/issues/133) issue.
> 
> You could configure `TcpListener` in the outer context. That does not belong to this library.

I don't believe this is possible, because the TLS server uses the TlsAcceptor object and not a TcpListener.

As this is an open source project, I will give my honest opinion, which can be taken into account or not.

I don't think it's a good implementation to configure the TcpListener in the external context. Any library that supports TCP or TLS has listeners configured within the library. For example, a REST API usually supports http and https in the library itself, such as Rocket (https://rocket.rs/).

In the rocket case, you just build a server: 
rocket::build().mount("/", routes![hello])

and specify the port to listen in a configuration file rocket.toml:
[debug]
port = 8443

In the same file you specify the certificates and keys after in this label:
[global.tls]
certs = 
key = 

Server operation is completely internal to the library, you just write the function to the routes.

I also don't understand this approach of saying that Modbus Security doesn't belong in a library that implements the Modbus protocol. The use of modbus TCP is only recommended for legacy systems. There is a commercial library that implements the entire Modbus protocol, but it uses a commercial license. It's called rodbus (https://docs.rs/rodbus/latest/rodbus/). You can see in that library that the Listener is created inside the library. 

   // spawn a server to handle connections onto its own task
   // if the handle _server is dropped, the server shuts down
   let _server = rodbus::server::spawn_tcp_server_task(
       1,
       SocketAddr::from_str("127.0.0.1:502")?,
       map,
       AddressFilter::Any,
       DecodeLevel::default(),
   ).await?;

You just specify the SocketAddr, as in the previous version of tokio-modbus.

Although I preferred to use the library created by you, I don't understand why you have been so reticent in accepting relevant contributions from the community, like the pull request "Fix/tcp delay causing composite tcp packets", which is extremely important to comply with the modbus specification.

I have a modbus TLS implementation based on this new stable tcp server. I can make a pull request as soon as version 0.8 comes out with the new server. Let me know if you are interested. However, if you are not interested in giving native support to modbus security in tokio-modbus, it would be interesting to make that clear. If so, I can work on a derived implementation, since tokio-modbus uses an MIT license.

I would really like to contribute with this project, but it has been very difficult to get my ideas accepted. Issues or pull requests are basically rejected with few comments. It is important to provide clear responses to issues, as documentation is still limited.

I really think the project is very interesting. Congrats. The only problem is the documentation and examples. After better understanding of the project's proposal, I was able to solve most of the problems I had with the server. If you are interested, I can work on a more complete server example with better documentation.

I would strongly like to contribute to the project, even if it were just discussing some ideias. Finally, I really believe that the tokio-modbus library can be a reference for many products based on modbus, rust and open source. It is a great projetct using the full power of the rust language.

Best Regards,
Gustavo

Issue #112, opened 2022-04-11
Maximum number of input registers read


chengchangwu:
I've found that I can read at most 126 input registers with `read_input_registers()`. But why 126? Which part of the source code should I read.

chengchangwu:
But I can read 126 contiguous registers. one more thsn the spec.

uklotzde:
This bug affects multiple function codes. Currently there are no semantic, client-side parameters according to the Modbus protocol validations in place. If a parameter value could technically be serialized it is accepted, even if the Modbus protocol specifies more restrictive, semantic bounds. Ideally servers should reject such non-standard requests.

Not a severe bug as clients are responsible for sending valid requests and servers are advised to reject invalid requests. Enforcing strict standard behavior by the library may even prevent using custom, vendor-specific extensions.

uklotzde:
Not a bug, it's a "feature".

If those additional conformance checks are implemented then they should be hidden behind a feature flag. Not everyone might need or want them.

uklotzde:
This affects both the client and the server.

Issue #40, opened 2020-01-31
Support for function 43 / 14 Read Device Identification


psztoch:

flosse:
@psztoch Would you mind to implement this?

ptasz3k:
I can try it. I'm learning rust, using modbus professionally and probably will implement that function in my c and c# modbus libs. I think that it will be good exercise.

Any hints where should i start?

flosse:
> I can try it.

awesome!

> Any hints where should i start?

You can find an example described on page 43 in the [Modbus Application Protocol Specification (PDF)](http://modbus.org/docs/Modbus_Application_Protocol_V1_1b3.pdf).

![Screenshot_2020-04-20 MODBUS Application Protocol 1 1 b - Modbus_Application_Protocol_V1_1b3 pdf](https://user-images.githubusercontent.com/276043/79773995-7b645d80-8332-11ea-9949-195118b09d52.png)

So here are some possible steps to implement it:

1. Define the Request & Response
2. Write test(s) with the given example
3. Implement the processing algorithm
4. Extend the public API

gustavowd:
Hi.
Any chance to merge the read device identification function to the code?
Its stall since 2020.

XavierJEsteve:
Hello,
I will begin investigating this as I learn the package. Motivated to do so as I am new to ICS  and learning Rust. I will post questions here.

PR #160, opened 2023-03-05
Encoding of a request into bytes might fail

uklotzde:
- Prerequisite for #40/#58 and #112.
- Allowing this conversion to fail does not affect the public API


flosse (approved):
LGTM!

PR #162, opened 2023-03-06
Possible approach for Modbus Security client and server

gustavowd:
This is an idea of possible implementation of the modbus secondity specification. It can be used to create a version more appropriate by project maintenance.


gustavowd:
I don´t have complete sure of how to do that, but seems to me that is just do something like that:
use tokio::io::split;

let (mut reader, mut writer) = split(stream);

But i don´t know how to change the tokio-modbus code in order to use that approach.

PR #163, opened 2023-03-07
example: tcp server additions for read and write, error handling

cdbennett:
Attempting to expand the TCP server example by adding support for holding register read/write support, plus demonstrating error handling on both client and server side.

Need help to figure out how to send specific Modbus exception responses on the server side, and how to identify the exception code on the client side.  For instance, in this example we test accessing a register address that is not supported by the server, so the server should respond with the Modbus exception code `IllegalDataAddress`.

### DIFF
--- a/examples/tcp-server.rs
+++ b/examples/tcp-server.rs
@@ -65,10 +65,10 @@ impl tokio_modbus::server::Service for ExampleService {
             _ => {
                 println!("SERVER: Exception::IllegalFunction - Unimplemented function code in request: {req:?}");
                 // TODO: We want to return a Modbus Exception response `IllegalFunction`. https://github.com/slowtec/tokio-modbus/issues/165
-                return future::ready(Err(std::io::Error::new(
+                future::ready(Err(std::io::Error::new(
                     std::io::ErrorKind::AddrNotAvailable,
-                    format!("Unimplemented function code in request"),
-                )));
+                    "Unimplemented function code in request".to_string(),
+                )))
             }
         }
     }
@@ -179,7 +179,7 @@ async fn client_context(socket_addr: SocketAddr) {
             assert_eq!(response, vec![1234, 5678]);
 
             println!("CLIENT: Writing 2 holding registers...");
-            ctx.write_multiple_registers(0x01, &vec![7777, 8888])
+            ctx.write_multiple_registers(0x01, &[7777, 8888])
                 .await
                 .unwrap();
 
--- a/examples/tcp-server.rs
+++ b/examples/tcp-server.rs
@@ -1,18 +1,29 @@
 // SPDX-FileCopyrightText: Copyright (c) 2017-2023 slowtec GmbH <post@slowtec.de>
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
-//! TCP server example
+//! # TCP server example
+//!
+//! This example shows how to start a server and implement basic register
+//! read/write operations.
 
-use std::{net::SocketAddr, time::Duration};
+use std::{
+    collections::HashMap,
+    net::SocketAddr,
+    sync::{Arc, Mutex},
+    time::Duration,
+};
 
 use futures::future;
 use tokio::net::TcpListener;
 
 use tokio_modbus::{prelude::*, server::tcp::Server};
 
-struct Service;
+struct ExampleService {
+    input_registers: Arc<Mutex<HashMap<u16, u16>>>,
+    holding_registers: Arc<Mutex<HashMap<u16, u16>>>,
+}
 
-impl tokio_modbus::server::Service for Service {
+impl tokio_modbus::server::Service for ExampleService {
     type Request = Request;
     type Response = Response;
     type Error = std::io::Error;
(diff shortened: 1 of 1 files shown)


examples/tcp-server.rs
```diff
@@ -1,35 +1,144 @@
...
+    registers: &mut HashMap<u16, u16>,
+    addr: u16,
+    values: &[u16],
+) -> Result<(), std::io::Error> {
+    for (i, value) in values.iter().enumerate() {
+        let reg_addr = addr + i as u16;
+        if let Some(r) = registers.get_mut(&reg_addr) {
+            *r = *value;
+        } else {
+            return Err(std::io::Error::new(
```
uklotzde, line 132: Also add the TODO about returning IllegalDataAddress here?

examples/tcp-server.rs
```diff
@@ -1,35 +1,144 @@
...
+    addr: u16,
+    cnt: u16,
+) -> Result<Vec<u16>, std::io::Error> {
+    let mut response_values = vec![0; cnt.into()];
+    for i in 0..cnt {
+        let reg_addr = addr + i;
+        if let Some(r) = registers.get(&reg_addr) {
+            response_values[i as usize] = *r;
+        } else {
+            // TODO: We want to return a Modbus Exception response `IllegalDataAddress`.
```
uklotzde, line 108: We should probably create and issue and mark it as a question. Adding a link to this issue here would help to keep track of it. Could then be resolved independently after merging.

uklotzde (commented):
Thank you! Very helpful addition for getting started. And it also revealed some gaps.

Only some minor comments.

examples/tcp-server.rs
uklotzde: Here is the issue: #165

uklotzde: Let's finish this PR first and then discuss any details in a follow-up PR. Easier to reason about with some working and verified code at hand ;)

cdbennett: Added the todo and linked question issue. Also created client side question #169

ladvoc:
This updated example for the TCP server has clarified many questions I had. However, one question I still have is how to store an `Arc` cloned from somewhere else when the service is created. For instance, is it possible to clone an existing `Arc<Mutex<HashMap<u16, u16>>` to store in the `input_registers` property when the `Service` is created? The reason for doing something like this is to enable the contents of the hash map to be fetched from some external source on another thread. Based on the requirements for `NewService`, I cannot figure out a way to accomplish this. I am relatively new to Rust, so I apologize if there is something obvious I have overlooked.

gustavowd:
> 

Insted o call the new() function of the Service, you can create the data objects in the main function and clone it to your threads.
For example:

#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    let socket_addr = "127.0.0.1:5502".parse().unwrap();

    // Insert some test data as register values.
    let mut input_registers = HashMap::new();
    input_registers.insert(0, 1234);
    input_registers.insert(1, 5678);
    let mut holding_registers = HashMap::new();
    holding_registers.insert(0, 10);
    holding_registers.insert(1, 20);
    holding_registers.insert(2, 30);
    holding_registers.insert(3, 40);
    let ir = Arc::new(Mutex::new(input_registers));
    let hr = Arc::new(Mutex::new(holding_registers));

    tokio::select! {
        _ = server_context(socket_addr,Arc::clone(&ir), Arc::clone(&hr)) => unreachable!(),
        _ = another_thread_context(socket_addr,Arc::clone(&ir)) => println!("Exiting"),
    }

    Ok(())
}

async fn server_context(socket_addr: SocketAddr, ir: Arc<Mutex<HashMap<u16, u16>>>, hr: Arc<Mutex<HashMap<u16, u16>>>) -> anyhow::Result<()> {
    println!("Starting up server on {socket_addr}");
    let listener = tokio_modbus::server::tls::listener(socket_addr, 1).unwrap();
    let server = Server::new(listener);
    let new_service = |_socket_addr| Some(Service {input_registers: Arc::clone(&ir), holding_registers:Arc::clone(&hr)});
    let on_process_error = |err| {
        eprintln!("{err}");
    };
    server.serve(&new_service, on_process_error).await?;
    Ok(())
}

fn client_context(uart_addr: &str,  ir: Arc<Mutex<HashMap<u16, u16>>>) {
...

Or implement a new() function of the Service that receive the data instead of create inside the new function.

gustavowd:
I think this library is great, just lacking some better examples and documentation. It will came with time. I made myself an example where holding registers are data structures with several different data such floating point and strings which uses a from trair to convert the scructs to Vec<u16>. This structs generate data in different threads using something like I suggest previouslly.

gustavowd:
> 
How to implement service to create the registers HashMap in the main function:

impl Service {
    fn new(ir: Arc<Mutex<HashMap<u16, u16>>>, hr: Arc<Mutex<HashMap<u16, u16>>>) -> Self {
        Self {
            input_registers: ir,
            holding_registers: hr,
        }
    }
}

uklotzde (approved):

PR #164, opened 2023-03-09
Use AsyncRead/AsyncWrite traits for generic TCP implementation

### LINKED ISSUE #119: Make TCP accept AsyncRead + AsyncWrite
Hi!
Would you accept a PR with reworks Modbus TCP to allow to use AsyncRead + AsyncWrite?
As a TcpStream has too  AsyncRead + AsyncWrite so that shouldn't be hard to do.
Of course we could keep the current `connect` and `connect_slave` and just add `connect_slave_from_transport` or something like that.
My use case for that:
- doing some connecting setup before starting working with tokio-modbus, I have connections where I first need to send a 'login' packet to have access to modbus
- running modbus-tcp over udp, not a typical use case but I sometimes have to do that

uklotzde:
- Fixes #119
- See the changelog for the many changes

Going through the code revealed inconsistencies, bugs, and the unneeded use of `async` and `Result`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - TCP Server: Stabilize "tcp-server" feature
 - TCP Client: Allow to attach a generic transport layer, e.g. for TLS support
 - RTU Server: Remove `NewService` trait and pass instance directly
+- Fix (sync): No timeout while establishing serial RTU connections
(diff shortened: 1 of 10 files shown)


flosse (approved):
nice :+1:

PR #166, opened 2023-03-09
TLS client based on new transport attach function

gustavowd:


pki/generate_cert_and_key.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
uklotzde, line 1: This file with very opinionated and even contains personalized details does not belong to the repo.

A link to external documentation on how to generate the certificates in the example would be more suitable and sufficient. This is not specific for Modbus.

Cargo.toml
```diff
@@ -37,6 +37,11 @@ tokio-serial = { version = "5.4.4", optional = true, default-features = false }
 tokio-util = { version = "0.7.7", default-features = false, features = [
     "codec",
 ] }
+rustls-pemfile = {version="1.0.2", optional = true}
```
uklotzde, line 40: These dependencies are nedeed only for the exampke.

Cargo.toml
```diff
@@ -50,9 +55,10 @@ tokio = { version = "1.26.0", default-features = false, features = [
 tokio-serial = { version = "5.4.4", default-features = false }
 
 [features]
-default = ["rtu", "tcp", "rtu-server", "tcp-server"]
+default = ["rtu", "tcp", "rtu-server", "tcp-server", "tls"]
 rtu = ["futures-util/sink"]
 tcp = ["tokio/net", "futures-util/sink"]
+tls = ["dep:rustls-pemfile", "dep:tokio-rustls", "dep:webpki-roots", "dep:pkcs8", "dep:pem"]
```
uklotzde, line 61: This feature is unneeded.

uklotzde (changes requested):

Cargo.toml
gustavowd: Ok. All suggestions accepted.

examples/tls-client.rs
```diff
@@ -0,0 +1,122 @@
...
+    } else {
+        let content = std::fs::read(path).unwrap();
+        let mut iter = pem::parse_many(content)
+            .unwrap()
+            .into_iter()
+            .filter(|x| x.tag == expected_tag)
+            .map(|x| x.contents);
+
+        match iter.next() {
+            Some(key) => match password {
```
uklotzde, line 45: Why not `match (key, password)`?

examples/tls-client.rs
```diff
@@ -0,0 +1,122 @@
...
+                Some(password) => {
+                    //println!("{:?}", key);
```
uklotzde, line 47: Commented out code is bad. Either delete or comment why we must keep it.

examples/tls-client.rs
```diff
@@ -0,0 +1,122 @@
...
+
+use pkcs8::der::Decode;
+use rustls_pemfile::{certs, pkcs8_private_keys};
+use tokio_rustls::rustls::{self, Certificate, OwnedTrustAnchor, PrivateKey};
+use tokio_rustls::{webpki, TlsConnector};
+
+fn load_certs(path: &Path) -> io::Result<Vec<Certificate>> {
+    certs(&mut BufReader::new(File::open(path)?))
+        .map_err(|_| io::Error::new(io::ErrorKind::InvalidInput, "invalid cert"))
+        .map(|mut certs| certs.drain(..).map(Certificate).collect())
```
uklotzde, line 23: Why not `certs.into_iter()`?

examples/tls-client.rs
```diff
@@ -0,0 +1,122 @@
...
+fn load_keys(path: &Path, password: Option<&str>) -> io::Result<Vec<PrivateKey>> {
+    let expected_tag = match &password {
+        Some(_) => "ENCRYPTED PRIVATE KEY",
+        None => "PRIVATE KEY",
+    };
+
+    if expected_tag.eq("PRIVATE KEY") {
+        pkcs8_private_keys(&mut BufReader::new(File::open(path)?))
+            .map_err(|_| io::Error::new(io::ErrorKind::InvalidInput, "invalid key"))
+            .map(|mut keys| keys.drain(..).map(PrivateKey).collect())
```
uklotzde, line 35: keys.into_iter()?

gustavowd: I don't master rust enough to understand these optimizations you are proposing.  As you already said, we can proceed with the pull request and after the code is added to the repository, we can suggest improvements with a new pull request.  What do you think?

examples/tls-client.rs
gustavowd: Most of this code for load certs and keys is from an example of the official git of the tokio-rustls project.
[https://github.com/tokio-rs/tls/blob/master/tokio-rustls/examples/server/src/main.rs](https://github.com/tokio-rs/tls/blob/master/tokio-rustls/examples/server/src/main.rs).

So, I believe we can leave it like this.

uklotzde: If you copied code than please at least add a comment with a link to it. Since the original source is only an example and the licenses are compatible I don't expect any complaints. Nevertheless respecting copyrights is always required.

examples/tls-client.rs
```diff
@@ -0,0 +1,111 @@
+// SPDX-FileCopyrightText: Copyright (c) 2017-2023 slowtec GmbH <post@slowtec.de>
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+//! Asynchronous TLS client example
```
uklotzde, line 4: Please add a paragraph with a remark including the link to the original sources you copied the code from here.

examples/tls-client.rs
gustavowd: I've never seen anyone consider copying code without respecting copyright when using code that exemplifies how to use a library from another project. But if you consider that this is necessary, I see no problem in making this note.

examples/tls-client.rs
gustavowd: Done!

examples/tls-client.rs
uklotzde: You neither resolved this issue nor answered my comment. Please do not resolve issues yourself if you simply did nothing and decided to silently ignore it.

Anyway. Then let's leave it as. Though the printed messages do not reflect the actual error as you might have noticed.

gustavowd: Sorry, this one i marked by mistake.

uklotzde (approved):
Thanks.

The shown implementation is still subpar and dangerous with all the unwraps. But since this is only simple client example I don't care enough and don't want to spend more time on this review.

PR #168, opened 2023-03-10
Possible solution for the server tls implementation

gustavowd:
The only problem is that TlsAcceptor do not implement the Debug trait.
Despite of that, I think this is the cleanner solution to implement both tcp and tls server with the same code.

### DIFF
--- a/src/server/tcp.rs
+++ b/src/server/tcp.rs
@@ -11,7 +11,7 @@ use futures_util::{future::FutureExt as _, sink::SinkExt as _, stream::StreamExt
 use socket2::{Domain, Socket, Type};
 use tokio::{
     io::{AsyncRead, AsyncWrite},
-    net::TcpListener,
+    net::{TcpListener, TcpStream},
 };
 use tokio_util::codec::Framed;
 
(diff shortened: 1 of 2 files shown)


src/server/tcp.rs
```diff
@@ -29,17 +33,17 @@ pub trait BindSocket {
 
     async fn bind_socket(addr: SocketAddr) -> Result<Socket, Self::Error>;
 }
-
-#[derive(Debug)]
+#[allow(missing_debug_implementations)]
 pub struct Server {
     listener: TcpListener,
+    tls_acceptor: Option<TlsAcceptor>,
```
uklotzde, line 39: I don't think that this is a good idea. Instead a generic solution is required for accepting connections.

gustavowd: I couldn't find another way to generalize the server, because the tokio-rustls based server needs a TcpListener and a TlsAcceptor.  I believe that it is unlikely that it will be possible to pass only one of the objects and be able to create the server.  Anyway, you can try to help think about the solution.  That way it works perfectly with minimal modifications to the current library.  However, I think it's unlikely that it's possible to generalize the code to the point where you don't need to include tokio-rustls as a dependency to create the server.

gustavowd: Furthermore, even if only the TlsAcceptor were passed as a parameter, it needs to be cloned to copy the certificate information for each connection.  this is not provided for in the TCP server.  They are different transport layers, it will be very difficult to fully abstract tls to the point of being able to use the same server as TCP.

Issue #111, opened 2022-02-25
Failure to Recover from TCP Re-transmission


Carter12s:
Totally possible this is user error, or some misconfiguration, but it looks like a failure in either tokio modbus or mio to handle TCP Re-transmissions correctly. We have two client processes (one reading, one writing) talking to a remote server (ABB VFD) in a harsh EM environment where there are a reasonable number of low level transmission failures.

We see a 100% repeatable pattern in network traffic of:

1. Request is sent, no response is received
2. After ~200ms a TCP Retransmission is sent from client to server
3. Immediately server responds with "TCP Previous segment not captured"
4. Client does not ack the "TCP Previous Segment not captured message"
5. After ~800ms a TCP Retransmission is sent from server to client
6. Request then fails

This may be an issue with how the server device is responding to the re-transmission, I'm unclear on the what the correct low-level responses should be in the situation, but its repeatable that whenever a TCP Re-tranmission occurs the request is never successfully recovered by TCP's retry mechanisms.

It should be noted that while the ~1 second delay from TCP retries is occurring, the other client is still successfully communicating with the device.

Client System Details:
Ubuntu 20.04, Rust 1.58.1, Tokio-Modbus 0.5.1

Wireshark screenshot of events:
![image](https://user-images.githubusercontent.com/8097919/155767066-6fd6260e-073b-45bf-8936-e97457057b6c.png)

PR #172, opened 2023-03-13
fix tcp retransmission by reading up to 10 frames

chenxinghua:


chenxinghua:
when **verify_response_header**  failed, try reading more frames

src/service/tcp.rs
```diff
@@ -75,15 +76,27 @@ where
...
 
-        match res_adu.pdu {
-            ResponsePdu(Ok(res)) => verify_response_header(req_hdr, res_adu.hdr).and(Ok(res)),
-            ResponsePdu(Err(err)) => Err(Error::new(ErrorKind::Other, err)),
+            match res_adu.pdu {
+                ResponsePdu(Ok(res)) => if let Err(e) = verify_response_header(req_hdr, res_adu.hdr) {
+                    if retries >= MAX_RETRIES {
+                        return Err(e);
+                    }
+                    retries += 1;
```
uklotzde, line 93: The error should at least be logged instead of silently ignoring it.

src/service/tcp.rs
```diff
@@ -18,6 +18,7 @@ use crate::{
 };
 
 const INITIAL_TRANSACTION_ID: TransactionId = 0;
+const MAX_RETRIES: usize = 10;
```
uklotzde, line 21: Max retries of what? Why 10? If this is an arbitrary number then it should better be configurable instead of a hard-coded constant.

A comment with references why this retry algorithm is needed is also required. Either by documenting the constant or the corresponding configuration option.

uklotzde (changes requested):
Thanks for taking care. I have left some comments with questions.

src/service/tcp.rs
chenxinghua:
```suggestion
                    if retries >= MAX_RETRIES {
                        return Err(e);
                    }
                    retries += 1;
                    log::info!("Try to recover from header mismatch: expected/request = {:?}, actual/response = {:?}, retries = {:?}", req_hdr, res_adu.hdr, retries);
```

log like this is acceptable?

src/service/tcp.rs
chenxinghua: Let **MAX_RETRIES** be configurable is more suitable. Can you give me some advice? I have not found how to add configurable code in this crate.

The retry algorithm is inspired by **libmodbus**'s error recovery mode, see [https://libmodbus.org/reference/modbus_set_error_recovery/](url) for details.

> When MODBUS_ERROR_RECOVERY_PROTOCOL is set, a sleep and flush sequence will be used to clean up the ongoing communication, this can occurs when the message length is invalid, the TID is wrong or the received function code is not the expected one. The response timeout delay will be used to sleep.

src/service/tcp.rs
uklotzde: Please modify the code directly instead of replying to comments. Commits can be squashed later before merging.

src/service/tcp.rs
uklotzde: If this parameter should be configurable then make a proposal. Otherwise use a constant and comment (in the code) why.

src/client/tcp.rs
```diff
@@ -26,6 +26,20 @@ pub async fn connect_slave(socket_addr: SocketAddr, slave: Slave) -> Result<Cont
...
 
+/// Connect to a physical, broadcast, or custom Modbus device,
+/// and can enable recover strategy when modbus header mismatch,
+/// when one response frame isn't arrived in specified time due
+/// to tcp retransmission, in the next reading or writing previous
+/// reponse frame will arrived, with this strategy the mismatched
+/// response frame can be discarded. This strategy is inspired by
+/// libmodbus's error recovery: 
+/// https://libmodbus.org/reference/modbus_set_error_recovery/
+pub async fn connect_slave_recover(socket_addr: SocketAddr, slave: Slave, max_recover_retries: usize) -> Result<Context, Error> {
```
uklotzde, line 37: The number of factory functions for creating a client context gets confusing with all the possible combinations. I guess switching to a builder pattern would be more appropriate, although this is bigger tasks apart from this PR.

@flosse Your thoughts?

src/client/tcp.rs
```diff
@@ -26,6 +26,20 @@ pub async fn connect_slave(socket_addr: SocketAddr, slave: Slave) -> Result<Cont
     Ok(context)
 }
 
+/// Connect to a physical, broadcast, or custom Modbus device,
```
uklotzde, line 29: - Please try to use short sentences
- The initial line or sentence should be the summary, followed by one or more paragraphs with details.

But let's wait until we decided if you use an implicit constant for the time being before we are switching to a configurable parameter when introducing the builder pattern.

src/client/tcp.rs
```diff
@@ -26,6 +26,20 @@ pub async fn connect_slave(socket_addr: SocketAddr, slave: Slave) -> Result<Cont
...
+/// and can enable recover strategy when modbus header mismatch,
+/// when one response frame isn't arrived in specified time due
+/// to tcp retransmission, in the next reading or writing previous
+/// reponse frame will arrived, with this strategy the mismatched
+/// response frame can be discarded. This strategy is inspired by
+/// libmodbus's error recovery: 
+/// https://libmodbus.org/reference/modbus_set_error_recovery/
```
uklotzde, line 36: When following the link I don't see any connection between the recovery modes offered by libmodus and this parameter. They don't even mention any retires? This is more confusing than it helps.

uklotzde: I quickly checked the rodbus implementation and didn't find anything similar.

Please elaborate why this retry is needed and desired. We should not introduce common parameters that only solve a particular use case.

uklotzde:
I am concerned that this might only be a quick band aid for a more general issue, namely that response messages could arrive out of order. Is this even possible or allowed over a single connection? I am not deeply familiar with the Modbus specifications.

src/client/tcp.rs
chenxinghua: In **libmodbus**, discard ready frames when transcation id mismatch and error recover strategy enabled.

```c
        rc = ctx->backend->pre_check_confirmation(ctx, req, rsp, rsp_length);
        if (rc == -1) {
            if (ctx->error_recovery & MODBUS_ERROR_RECOVERY_PROTOCOL) {
                _sleep_response_timeout(ctx);
                modbus_flush(ctx);
            }
            return -1;
        }
``` 

I assume that, first several (1 or 2) frames are mismatch, the ok frame may in the remaining.

For situation in #111 discard one frame, the remaining can work.

chenxinghua:
any further suggestion? should I let MAX_RETRIES constant hardcoded for accepted. @uklotzde

uklotzde:
Please re-open when done.

PR #173, opened 2023-03-13
Possible approach to implement the tls server

gustavowd:
I believe that mantainers intended to not include tokio-rustls as the project dependencies. However, i could not think in any solution to implement the accept_tls_connection() function due to the RequestAdu and OptionalResponsePdu, which are private from the frame module. Any suggestion of how to solve this issue?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,11 @@ tokio = { version = "1.26.0", default-features = false, features = [
     "time",
 ] }
 tokio-serial = { version = "5.4.4", default-features = false }
+rustls-pemfile = "1.0.2"
+tokio-rustls = "0.23.4"
+webpki-roots = "0.22.6"
+pkcs8 = { version = "0.10.1", features = ["encryption", "pem", "std"] }
+pem = "1.1.1"
 
 [features]
 default = ["rtu", "tcp", "rtu-server", "tcp-server"]
@@ -110,3 +115,8 @@ required-features = ["tcp"]
 name = "tcp-server"
 path = "examples/tcp-server.rs"
 required-features = ["tcp-server"]
+
+[[example]]
+name = "tls-client"
+path = "examples/tls-client.rs"
+required-features = ["tcp"]
--- a/examples/tls-client.rs
+++ b/examples/tls-client.rs
@@ -1,0 +1,114 @@
+// SPDX-FileCopyrightText: Copyright (c) 2017-2023 slowtec GmbH <post@slowtec.de>
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+// load_certs() and particially load_keys() functions were copied from an example of the tokio tls library, available at:
+// https://github.com/tokio-rs/tls/blob/master/tokio-rustls/examples/server/src/main.rs
+
+//! Asynchronous TLS client example
+use tokio::net::TcpStream;
+
+use std::{
+    fs::File,
+    io::{self, BufReader},
+    net::SocketAddr,
+    path::Path,
+    sync::Arc,
+};
+
+use pkcs8::der::Decode;
+use rustls_pemfile::{certs, pkcs8_private_keys};
+use tokio_rustls::rustls::{self, Certificate, OwnedTrustAnchor, PrivateKey};
+use tokio_rustls::{webpki, TlsConnector};
+
+fn load_certs(path: &Path) -> io::Result<Vec<Certificate>> {
+    certs(&mut BufReader::new(File::open(path)?))
+        .map_err(|_| io::Error::new(io::ErrorKind::InvalidInput, "invalid cert"))
+        .map(|mut certs| certs.drain(..).map(Certificate).collect())
+}
+
+fn load_keys(path: &Path, password: Option<&str>) -> io::Result<Vec<PrivateKey>> {
+    let expected_tag = match &password {
+        Some(_) => "ENCRYPTED PRIVATE KEY",
+        None => "PRIVATE KEY",
+    };
+
+    if expected_tag.eq("PRIVATE KEY") {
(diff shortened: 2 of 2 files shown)


gustavowd:
I think i found a way to create the OnConnected() function without an accept_tls_connection() function.

examples/tls-server.rs
```diff
@@ -0,0 +1,311 @@
...
+        let certs = load_certs(cert_path).unwrap();
+        let mut keys = load_keys(key_path, None).unwrap();
+        let config = rustls::ServerConfig::builder()
+            .with_safe_defaults()
+            .with_no_client_auth()
+            .with_single_cert(certs, keys.remove(0))
+            .map_err(|err| io::Error::new(io::ErrorKind::InvalidInput, err)).unwrap();
+        let acceptor = TlsAcceptor::from(Arc::new(config));
+
+        let service = Some(ExampleService::new());
```
uklotzde, line 223: Enclosing in `Some` here and later using `map()` is pointless and confusing. Please simplify the code.

examples/tls-server.rs
```diff
@@ -0,0 +1,311 @@
...
+
+async fn server_context(socket_addr: SocketAddr) -> anyhow::Result<()> {
+    println!("Starting up server on {socket_addr}");
+    let listener = TcpListener::bind(socket_addr).await?;
+    let server = Server::new(listener);
+
+    let on_connected = |stream, _socket_addr| async move {
+        let cert_path = Path::new("./pki/server.pem");
+        let key_path = Path::new("./pki/server.key");
+        let certs = load_certs(cert_path).unwrap();
```
uklotzde, line 214: Using `unwrap()` and panicking on error in functions that return a `Result` is considered bad practice.

uklotzde (changes requested):

examples/tls-server.rs
```diff
@@ -0,0 +1,311 @@
...
+        Some(_) => "ENCRYPTED PRIVATE KEY",
+        None => "PRIVATE KEY",
+    };
+
+    if expected_tag.eq("PRIVATE KEY") {
+        pkcs8_private_keys(&mut BufReader::new(File::open(path)?))
+            .map_err(|_| io::Error::new(io::ErrorKind::InvalidInput, "invalid key"))
+            .map(|mut keys| keys.drain(..).map(PrivateKey).collect())
+    } else {
+        let content = std::fs::read(path).unwrap();
```
uklotzde, line 47: unwrap() instead of returning an error again

examples/tls-server.rs
```diff
@@ -0,0 +1,311 @@
...
+    let on_connected = |stream, _socket_addr| async move {
+        let cert_path = Path::new("./pki/server.pem");
+        let key_path = Path::new("./pki/server.key");
+        let certs = load_certs(cert_path).unwrap();
+        let mut keys = load_keys(key_path, None).unwrap();
+        let config = rustls::ServerConfig::builder()
+            .with_safe_defaults()
+            .with_no_client_auth()
+            .with_single_cert(certs, keys.remove(0))
+            .map_err(|err| io::Error::new(io::ErrorKind::InvalidInput, err)).unwrap();
```
uklotzde, line 220: You first map the error but then call unwap?

examples/tls-server.rs
gustavowd: Done

examples/tls-server.rs
gustavowd: You are right. Now the error is returned as Result

examples/tls-server.rs
```diff
@@ -0,0 +1,309 @@
...
+        let mut keys = load_keys(key_path, None)?;
+        let config = rustls::ServerConfig::builder()
+            .with_safe_defaults()
+            .with_no_client_auth()
+            .with_single_cert(certs, keys.remove(0))
+            .map_err(|err| io::Error::new(io::ErrorKind::InvalidInput, err))?;
+        let acceptor = TlsAcceptor::from(Arc::new(config));
+
+        let service = ExampleService::new();
+        let stream = acceptor.accept(stream).await.unwrap();
```
uklotzde, line 221: unwrap()?

examples/tls-server.rs
```diff
@@ -0,0 +1,309 @@
...
+    };
+
+    if expected_tag.eq("PRIVATE KEY") {
+        pkcs8_private_keys(&mut BufReader::new(File::open(path)?))
+            .map_err(|_| io::Error::new(io::ErrorKind::InvalidInput, "invalid key"))
+            .map(|mut keys| keys.drain(..).map(PrivateKey).collect())
+    } else {
+        let content = std::fs::read(path)?;
+        let mut iter = pem::parse_many(content)
+            .unwrap()
```
uklotzde, line 46: The error should be transformed into an io::Error and propagated.

examples/tls-server.rs
```diff
@@ -0,0 +1,309 @@
...
+            .into_iter()
+            .filter(|x| x.tag == expected_tag)
+            .map(|x| x.contents);
+
+        match iter.next() {
+            Some(key) => match password {
+                Some(password) => {
+                    let encrypted = pkcs8::EncryptedPrivateKeyInfo::from_der(&key).unwrap();
```
uklotzde, line 54: Also these 2 unwraps.

examples/tls-server.rs
```diff
@@ -0,0 +1,309 @@
...
+            ));
+        }
+    }
+
+    Ok(())
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let socket_addr = "127.0.0.1:8802".parse().unwrap();
```
uklotzde, line 192: Replace `unwrap()` with `?`

uklotzde:
Please rebase and resolve the conflicts.

PR #174, opened 2023-03-14
Possible approach for the TLS server.

gustavowd:


uklotzde:
Why did you close the previous PR with my comments and instead reopened a new one???

gustavowd:
> Why did you close the previous PR with my comments and instead reopened a new one???

Sorry.  It wasn't intentional.  I've never done a rebase before.  When I tried to resolve the conflict the pull request disappeared.

uklotzde:
Pull requests can only be 

> > Why did you close the previous PR with my comments and instead reopened a new one???
> 
> Sorry. It wasn't intentional. I've never done a rebase before. When I tried to resolve the conflict the pull request disappeared.

Pull requests are only closed implicitly if you delete the underlying branch. Rebasing the branch locally and then force pushing should work as expected.

uklotzde (approved):
Thank you.

The client-side unwraps are ok for such a simple example. That's a different case than for the server-side which explicitly offers returning `Result`s in the framework code. The appropriate usage of those closure signatures should be demonstrated by an example.

PR #176, opened 2023-03-15
Use more restrictive clippy settings

uklotzde:


src/lib.rs
```diff
@@ -9,13 +9,13 @@
...
-#![warn(clippy::all)]
-#![warn(clippy::cast_lossless)]
-#![warn(clippy::explicit_deref_methods)]
-#![warn(clippy::explicit_into_iter_loop)]
-#![warn(clippy::explicit_iter_loop)]
-// TODO: Decorate functions with #[must_use]
-//#![warn(clippy::must_use_candidate)]
+#![warn(clippy::pedantic)]
+#![warn(clippy::clone_on_ref_ptr)]
+#![allow(clippy::wildcard_imports)]
```
flosse, line 14: My personal preference is not to use wildcard imports but if everyone else like it, I'm totally fine :wink:

flosse (approved):

src/lib.rs
uklotzde: :+1: Then we should append `// TODO` as for other temporarily allowed attributes.

PR #186, opened 2023-04-21
Max retries

### LINKED ISSUE #184: Infinite retry loop with heavy CPU load
I have tested tokio-modbus on a misbehaving Modbus RTU bus, which currently responds with an infinite stream of zeros to all read requests. This is not a theoretical situation, but can happen when a single device on the bus is misconfigured.
I don't expect tokio-modbus to recover from such situations, which is why I've encapsulated the `read_holding_registers` call in a `tokio::time::timeout` and drop the client session after every error/timeout.

However, tokio-modbus currently tries to recover from that and thereby enters an infinite loop without delay or even exponential backoff. My timeout kills that read operation after a few seconds without data, but in the meantime, it retries every millisecond and thereby causes heavy CPU load.

In particular, I'm talking about this loop:
https://github.com/slowtec/tokio-modbus/blob/b8bdb90453e0c6c7caf5e7eb0c0a7ee44a93f55a/src/codec/rtu.rs#L242-L263

`get_pdu_len` reads a zero byte here, therefore returns with an `Err` that is caught by the `or_else` branch. That branch calls `recover_on_error` to clear that zero byte, sets `retry` to `true`, and tries again immediately.

I'm open for any solution that adds a short delay here.
Either via a `retry_delay` parameter that is passed to `tokio::time::sleep` before every retry, or via a passed function that is called before every retry (allows the user to implement exponential backoff).

CC @flosse @uklotzde

ColinFinck:
As suggested in #184, I've fixed the infinite loop when reading from a broken Modbus RTU bus by introducing a maximum number of retries. When that number is reached, the code gives up and returns an error to the caller.

`MAX_RETRIES` is a constant that is set to 20 for now. Feel free to change that or make it configurable.

I had to edit other parts of the RTU code, which assumed that decoding can never fail. I took the opportunity to make them more readable and more consistent with the TCP code. Should get obvious when you check this PR commit by commit.

Let me know if this can go through or if there's anything I have to change.

Fixes #184

### DIFF
--- a/src/codec/rtu.rs
+++ b/src/codec/rtu.rs
@@ -244,20 +244,22 @@ where
     const MAX_RETRIES: usize = 20;
 
     for _i in 0..MAX_RETRIES {
-        match get_pdu_len(buf) {
-            Ok(pdu_len) => {
-                let Some(pdu_len) = pdu_len else {
-                    // Incomplete frame
-                    return Ok(None);
-                };
+        let result = get_pdu_len(buf).and_then(|pdu_len| {
+            let Some(pdu_len) = pdu_len else {
+                // Incomplete frame
+                return Ok(None);
+            };
 
-                return frame_decoder.decode(buf, pdu_len);
-            }
-            Err(err) => {
-                log::warn!("Failed to decode {} frame: {}", pdu_type, err);
-                frame_decoder.recover_on_error(buf);
-            }
+            frame_decoder.decode(buf, pdu_len)
+        });
+
+        if let Err(err) = result {
+            log::warn!("Failed to decode {pdu_type} frame: {err}");
+            frame_decoder.recover_on_error(buf);
+            continue;
         }
+
+        return result;
     }
 
     // Maximum number of retries exceeded.
--- a/src/codec/rtu.rs
+++ b/src/codec/rtu.rs
@@ -238,29 +238,28 @@ fn decode<F>(
 where
     F: Fn(&BytesMut) -> Result<Option<usize>>,
 {
-    // TODO: Transform this loop into idiomatic code
-    loop {
-        let mut retry = false;
-        let res = get_pdu_len(buf)
-            .and_then(|pdu_len| {
-                debug_assert!(!retry);
+    const MAX_RETRIES: usize = 20;
+
+    for _i in 0..MAX_RETRIES {
+        match get_pdu_len(buf) {
+            Ok(pdu_len) => {
                 if let Some(pdu_len) = pdu_len {
-                    frame_decoder.decode(buf, pdu_len)
+                    return frame_decoder.decode(buf, pdu_len);
                 } else {
                     // Incomplete frame
-                    Ok(None)
+                    return Ok(None);
                 }
-            })
-            .or_else(|err| {
+            }
+            Err(err) => {
                 log::warn!("Failed to decode {} frame: {}", pdu_type, err);
                 frame_decoder.recover_on_error(buf);
-                retry = true;
-                Ok(None)
-            });
-        if !retry {
-            return res;
+            }
         }
     }
+
+    // Maximum number of retries exceeded.
+    log::error!("Giving up to decode frame after {} retries", MAX_RETRIES);
+    Err(Error::new(ErrorKind::InvalidData, "Too many retries"))
 }
 
 impl Decoder for ClientCodec {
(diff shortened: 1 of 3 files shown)


src/codec/rtu.rs
```diff
@@ -36,42 +36,47 @@ impl FrameDecoder {
         buf: &mut BytesMut,
         pdu_len: usize,
     ) -> Result<Option<(SlaveId, Bytes)>> {
+        const CRC_BYTE_COUNT: usize = 2;
```
uklotzde, line 39: :+1:

src/codec/rtu.rs
```diff
@@ -36,42 +36,47 @@ impl FrameDecoder {
...
-                        }
-                        let slave_id = adu_buf.split_to(1)[0];
-                        let pdu_data = adu_buf.freeze();
-                        return Ok(Some((slave_id, pdu_data)));
-                    }
-                    Err(err) => Err(err),
-                },
-                Err(err) => Err(err),
+
+        if buf.len() < adu_len + CRC_BYTE_COUNT {
```
uklotzde, line 43: I am also an advocate of using early returns to keep the main path straight.

src/codec/rtu.rs
```diff
@@ -36,42 +36,47 @@ impl FrameDecoder {
...
+            // Incomplete frame
+            return Ok(None);
+        }
+
+        let mut adu_buf = buf.split_to(adu_len);
+        let crc_buf = buf.split_to(CRC_BYTE_COUNT);
+
+        // Read trailing CRC and verify ADU
+        match Cursor::new(&crc_buf)
```
uklotzde, line 52: Since the `Ok` branch doesn't carry any value we we could use any early return for the error case? Would be consistent with the early return above. On the other hand the current code verifies that the unit type and nothing else is returned on `Ok`. I am fine with both, your choice.

src/codec/rtu.rs
```diff
@@ -238,60 +243,51 @@ fn decode<F>(
...
-                Ok(None)
-            });
-        if !retry {
-            return res;
+            }
         }
     }
+
+    // Maximum number of retries exceeded.
+    log::error!("Giving up to decode frame after {} retries", MAX_RETRIES);
```
uklotzde, line 266: `{MAX_RETRIES}`

src/codec/rtu.rs
```diff
@@ -300,36 +296,28 @@ impl Decoder for ServerCodec {
...
-                } else {
-                    Ok(None)
-                }
-            })
-            .map_err(|_| {
-                // Decoding the transport frame is non-destructive and must
-                // never fail!
-                unreachable!();
-            })
+        if let Some((slave_id, pdu_data)) = self.decoder.decode(buf)? {
```
uklotzde, line 299: A `let-else` binding with an early return would improve readability.

uklotzde (commented):
Thank you very much! Just some minor comments/questions.

Please rebase on main, I have just fixed the unrelated pre-commit issue.

uklotzde:
While you are at it: Please add an entry to the changelog. I guess it is time to finally release this new version.

src/codec/rtu.rs
ColinFinck: I changed that to an early return. Saves us one level of indentation :)

src/codec/rtu.rs
ColinFinck: Changed. I refrained from doing that as I didn't know about your MSRV policy. The only hint was `edition = "2021"` indicating that I can use all features of Rust 1.56.
Capturing identifiers in format strings has increased the Minimum Supported Rust Version to 1.58.

src/codec/rtu.rs
ColinFinck: Changed. Like above, I refrained from introducing that earlier, because it's only supported since Rust 1.65.
Even worse, while it's supported, it's currently not picked up by `rustfmt`, so people will introduce their own coding styles around let-else. I've written it in a way that I consider meaningful, but a future `rustfmt` may reformat that.

Along with this change, I've added a line `rust-version = "1.65"` to `Cargo.toml` and added an entry to the changelog.

ColinFinck:
I've addressed all your comments.
If you still have that Serial-to-USB adapter, which originally made you skip errors and implement error recovery, please verify that my new code still works with that. Other than that, it should be good to go.

uklotzde (approved):

PR #189, opened 2023-05-02
Since modbus RTU is a request and response protocol, make sure the read buffer is always clear before sending another request

gustavowd:
I had a lot of problems with errors in the async RTU client. Erros such:
kind: InvalidData, error: "Too many retries"
Kind(BrokenPipe)

After that fix, never got any erros in the same setup.

### DIFF
--- a/src/service/rtu.rs
+++ b/src/service/rtu.rs
@@ -52,6 +52,9 @@ where
         let req_adu = self.next_request_adu(req, disconnect);
         let req_hdr = req_adu.hdr;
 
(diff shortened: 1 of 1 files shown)


uklotzde:
I am not deeply familiar with this code anymore and busy with other projects. Someone else might take a look and comment on the proposed changes.

ColinFinck:
Thanks for bringing this up! #198 has just been merged, which implements the same fix for RTU _and_ TCP.
Please try again with the latest revision of tokio-modbus.

PR #192, opened 2023-05-21
fix: Require tokio/rt feature for sync scenarios

flosse:


uklotzde:
Just fixing the issue without a regression test that triggers the dependency/feature error is brittle.

uklotzde:
`cargo check --no-default-features --features rtu-sync` fails.

Ideally, all or at least selected feature combinations should be checked by the CI builds. Using all possible combinations does not make much sense, but checking single features in isolation like demonstrated here would help.

uklotzde:
Another argument to avoid overusing features for crate extensions. Using separate `sync` and `server` crates would be more appropriate.

The `sync` stuff doesn't fit in here at all. Offering a _sync-over-async_ bridge layer is pointless for an _async_-focused crate IMHO.

flosse:
> The `sync` stuff doesn't fit in here at all. Offering a _sync-over-async_ bridge layer is pointless for an _async_-focused crate IMHO.

I agree!

uklotzde:
After force pushing GitHub noticed the new workflow action.

uklotzde (approved):

PR #193, opened 2023-05-22
Remove sync-over-async bridge layer

flosse:
This PR removes all sync features and implementations.
For use cases where a sync implementation is desired, an example based on `modbus-core` might be an idea.


uklotzde:
- The regression test in #194 failed as expected.
- Bugfix version could be released after merging this PR. I have updated both changelog and version number.

uklotzde:
Unfortunately, GitHub does not run the new workflow action here. #194 will show if it works after cherry picking the fix.

uklotzde:
Should be tracked by an issue, not a PR.

PR #195, opened 2023-05-22
Generalize impl around `Service` usage

### LINKED ISSUE #190: Generalize impl around `Service` usage
In previous versions of the crate, I could wrap my `Service` in an `Arc` to pass into `Server::serve`.  In the `0.8` release, I needed to create a new delegate type that basically accomplishes the same thing, i.e.,

```rs
struct MbDelegate(Arc<MbServer>);

impl Service for MbDelegate {
// ...

  fn call(&self, req: Self::Request) -> Self::Future { self.0.call(req) }
}
```

A suggestion to make this delegation unnecessary, create a blanket impl, such as:

```rs
impl<D> Service for D
where
  D: Deref,
  D::Target: Service + ?Sized
```

So that anything that derefs into a `Service` can be used.  As it stands, re-using a service (versus the example documentation where it appears a new service is created on every request), is kind of cumbersome and doesn't seem to make a lot of sense for a stateful server.

Halkcyon:
Resolves #190

### DIFF
--- a/src/server/service.rs
+++ b/src/server/service.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
 use std::future::Future;
+use std::ops::Deref;
 
 /// A Modbus server service.
 pub trait Service {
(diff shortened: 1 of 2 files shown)


uklotzde:
Side note: Next time you should better use a feature branch in your forked repo. Using upstream branches for development is discouraged. You probably need to reset your `main` branch later.

PR #196, opened 2023-05-26
fix: clear cache before call

rise0chen:


rise0chen:
#189

uklotzde (commented):
Thank you for picking this up. Please provide some context and more information, including why the approach proposed in #189 is insufficient.

- The code seems to duplicated. Could it be extracted into a (generic) function?
- Please add documentation to this function that explains why all this is needed, ideally with additional references.
- Is it necessary to consume single bytes one by one?
- The empty loop body is unhandy. Adding an appropriate comment inside it would improve the readability.

rise0chen:
## timeout

```rust
tokio::time::timeout(Duration::from_millis(100), context.read_coils(addr, len))
```

Maybe we can recv a expired data, after cancel a call, because of slave use more time to processing and causing timeout.
The expired data will be stored in the read buffer of transport `self.framed.get_mut()`, so we should clear it.

## transmission error

Maybe we can recv fault data， because of transmission error.
The fault data will be stored in the read buffer of framed `self.framed.read_buffer_mut()`, so we should clear it.

rise0chen:
Maybe we can use [tokio::io::AsyncReadExt::read_buf](https://docs.rs/tokio/latest/tokio/io/trait.AsyncReadExt.html#method.read_buf) to clear `self.framed.get_mut()`, but the length of buf is difficult to determine.

uklotzde:
> Maybe we can use [tokio::io::AsyncReadExt::read_buf](https://docs.rs/tokio/latest/tokio/io/trait.AsyncReadExt.html#method.read_buf) to clear `self.framed.get_mut()`, but the length of buf is difficult to determine.

We don't need to change the code if it is not possible. Adding a TODO comment that it would be desirable to read all available bytes at once is also a valid choice.

uklotzde:
> ## timeout
> 
> ```rust
> tokio::time::timeout(Duration::from_millis(100), context.read_coils(addr, len))
> ```
> 
> Maybe we can recv a expired data, after cancel a call, because of slave use more time to processing and causing timeout. The expired data will be stored in the read buffer of transport `self.framed.get_mut()`, so we should clear it.
> ## transmission error
> 
> Maybe we can recv fault data， because of transmission error. The fault data will be stored in the read buffer of framed `self.framed.read_buffer_mut()`, so we should clear it.

Please extract the code into a function with a concise name and add a comment with the details.

Questions asked during a review do not need to be answered in the PR. I am just a generic placeholder for any future developer who stumbles over this code. Ideally, the code should answer all these questions, or even better, is written in a way that these questions don't even come up in the reader's head.

uklotzde:
We also need a concise sentence for the CHANGELOG that summarizes the fix with its consequences and implications for users of this library. I wonder why no one complained about these serious issues earlier.

ColinFinck:
#198 has just been merged, which clears the buffer before each send in a simpler way.
Please try again with the latest tokio-modbus revision. Closing this PR - reopen if problems persist.

rise0chen:
```rust
use std::{net::SocketAddr, pin::Pin, time::Duration};

use futures::future;
use tokio::net::TcpListener;

use tokio_modbus::{
    prelude::*,
    server::tcp::{accept_tcp_connection, Server},
};

struct ExampleService {}

impl tokio_modbus::server::Service for ExampleService {
    type Request = Request;
    type Response = Response;
    type Error = std::io::Error;
    type Future =
        Pin<Box<dyn future::Future<Output = Result<Self::Response, Self::Error>> + Send + Sync>>;

    fn call(&self, req: Self::Request) -> Self::Future {
        match req {
            Request::ReadHoldingRegisters(_addr, cnt) => {
                let data = vec![0; cnt as usize];
                Box::pin(future::ready(Ok(Response::ReadHoldingRegisters(data))))
            }
            Request::WriteMultipleRegisters(addr, values) => Box::pin(async move {
                tokio::time::sleep(Duration::from_millis(500)).await;
                Ok(Response::WriteMultipleRegisters(addr, values.len() as u16))
            }),
            Request::WriteSingleRegister(addr, value) => Box::pin(future::ready(Ok(
                Response::WriteSingleRegister(addr, value),
            ))),
            _ => {
                println!("SERVER: Exception::IllegalFunction - Unimplemented function code in request: {req:?}");
                // TODO: We want to return a Modbus Exception response `IllegalFunction`. https://github.com/slowtec/tokio-modbus/issues/165
                Box::pin(future::ready(Err(std::io::Error::new(
                    std::io::ErrorKind::AddrNotAvailable,
                    "Unimplemented function code in request".to_string(),
                ))))
            }
        }
    }
}

impl ExampleService {
    fn new() -> Self {
        Self {}
    }
}

#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    let socket_addr = "127.0.0.1:5502".parse().unwrap();

    tokio::select! {
        _ = server_context(socket_addr) => unreachable!(),
        _ = client_context(socket_addr) => println!("Exiting"),
    }

    Ok(())
}

async fn server_context(socket_addr: SocketAddr) -> anyhow::Result<()> {
    println!("Starting up server on {socket_addr}");
    let listener = TcpListener::bind(socket_addr).await?;
    let server = Server::new(listener);
    let new_service = |_socket_addr| Ok(Some(ExampleService::new()));
    let on_connected = |stream, socket_addr| async move {
        accept_tcp_connection(stream, socket_addr, new_service)
    };
    let on_process_error = |err| {
        eprintln!("{err}");
    };
    server.serve(&on_connected, on_process_error).await?;
    Ok(())
}

async fn client_context(socket_addr: SocketAddr) {
    let task = async {
        // Give the server some time for starting up
        tokio::time::sleep(Duration::from_secs(1)).await;

        println!("CLIENT: Connecting client...");
        let mut ctx = tcp::connect(socket_addr).await.unwrap();

        let read_reg = tokio::time::timeout(
            Duration::from_millis(200),
            ctx.read_holding_registers(0x01, 2),
        )
        .await;
        println!("{:?}", read_reg); // Should success

        let write_mul = tokio::time::timeout(
            Duration::from_millis(200),
            ctx.write_multiple_registers(0x01, &[0]),
        )
        .await;
        println!("{:?}", write_mul); // Should timeout
        tokio::time::sleep(Duration::from_millis(500)).await; // wait finish

        let write_sig = tokio::time::timeout(
            Duration::from_millis(200),
            ctx.write_single_register(0x01, 0),
        )
        .await;
        println!("{:?}", write_sig); // Should success

        println!("CLIENT: Done.")
    };
    tokio::join!(task, tokio::time::sleep(Duration::from_secs(5)));
}
```

#198 not solve timeout.
You can test this example. `write_sig` will return error, if not clear `self.framed.get_mut()`

Issue #60, opened 2020-05-19
Poor recovery from malformed frames in RTU


MightyPork:
We have a modbus slave which implements modbus in the usual vendor way, i.e., wrong: It does not send CRC with error frames.

```
Req: 0A 01 04 A1 00 01 AC 63
Resp: 0A 81 02 <no_crc>
```

To debug this, I copied the crate to a local folder and added trace logging to the RTU decoder.

<details>
  <summary>Trace logs (long)</summary>

```
[2020-05-19T11:37:31.393Z TRACE tokio_util::codec::framed_read] attempting to decode a frame
[2020-05-19T11:37:31.393Z TRACE tokio_modbus::codec::rtu] ClientCodec::decode(b"\x01")
[2020-05-19T11:37:31.393Z TRACE tokio_modbus::codec::rtu] ResponseDecoder::decode(b"\x01")
[2020-05-19T11:37:31.393Z TRACE tokio_modbus::codec::rtu] --- decode b"\x01"
[2020-05-19T11:37:31.393Z TRACE tokio_modbus::codec::rtu] get_response_pdu_len b"\x01"
[2020-05-19T11:37:31.393Z TRACE tokio_modbus::codec::rtu] pdu len none
[2020-05-19T11:37:31.393Z TRACE tokio_modbus::codec::rtu] PDU len is None
[2020-05-19T11:37:31.393Z TRACE tokio_modbus::codec::rtu] no retry
[2020-05-19T11:37:31.393Z TRACE tokio_modbus::codec::rtu] decode resulted in -> None
[2020-05-19T11:37:31.393Z TRACE tokio_modbus::codec::rtu] frame is partial, ask for more bytes
[2020-05-19T11:37:31.393Z TRACE tokio_util::codec::framed_read] attempting to decode a frame
[2020-05-19T11:37:31.393Z TRACE tokio_modbus::codec::rtu] ClientCodec::decode(b"\x01\x83")
[2020-05-19T11:37:31.393Z TRACE tokio_modbus::codec::rtu] ResponseDecoder::decode(b"\x01\x83")
[2020-05-19T11:37:31.393Z TRACE tokio_modbus::codec::rtu] --- decode b"\x01\x83"
[2020-05-19T11:37:31.393Z TRACE tokio_modbus::codec::rtu] get_response_pdu_len b"\x01\x83"
[2020-05-19T11:37:31.393Z TRACE tokio_modbus::codec::rtu] len=Some(2)
[...]

PR #198, opened 2023-06-07
Clear rx buffer before call

MightyPork:
This helps with error recovery on unreliable physical connections.

An older version of this workaround was running for over 2 years now in several industrial installations

Relates to #60

### DIFF
--- a/src/service/rtu.rs
+++ b/src/service/rtu.rs
@@ -52,6 +52,8 @@ where
         let req_adu = self.next_request_adu(req, disconnect);
         let req_hdr = req_adu.hdr;
 
+        self.framed.read_buffer_mut().clear();
+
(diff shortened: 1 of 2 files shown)


MightyPork:
haha we now have three independent PRs trying to fix that 2020 issue.

- #196 seems like an extremely complicated way to do the same and frankly I don't understand the code
- #189 does the right thing, but misses the TCP service

ColinFinck:
@uklotzde Looks good to me! Can you approve the GitHub Actions workflows, so that we can have a last check before this gets merged?

This should indeed supersede #189 and #196

uklotzde:
@ColinFinck Thanks for stepping in again!

Please proceed as proposed ;)

Issue #169, opened 2023-03-10
Client: how to identify Modbus exception code sent by server in response


cdbennett:
How can user code using tokio-modbus as a Modbus client detect what specific exception code a server returned?

Two interrelated parts of this question:

1. How user code can distinguish between a recoverable error during a Modbus request, (e.g. IllegalDataAddress, and a connection failure, e.g. TCP connection closed by remote host, or a protocol framing error.)
2. How can user code identify which specific Modbus exception code was returned by the server. (E.g. IllegalDataAddress vs IllegalFunction)

In the case of Modbus exception responses, we don't want to close the Modbus connection. But if the network connection was lost, then user code needs to re-create a new tokio-modbus connection and start over.

Related: 
- https://github.com/slowtec/tokio-modbus/pull/163
- https://github.com/slowtec/tokio-modbus/issues/165

Stupremee:
Hey :wave: 

I'm really in the need of finding out what Modbus error I got from a slave for my application.
I noticed it's not possible using `tokio-modbus` yet (at leat without doing silly `err.to_string().contains("illegal data address")`, and co), and found this issue.

After looking through the code. I found a fast, simple way to get access to the exception from a modbus server. You just have to make the `ExceptionResponse` and `Exception` types publicy accessible. Then someone using this crate can try to downcast the `std::io::Error` to the `ExceptionResponse` type to extract the error.

cdbennett:
This issue is resolved, verified working now in tokio-modbus v0.14.0.

Example of issuing a command like:

```rust
    let values = read_registers(connection, start, count)
    .await
    .context("communication error")?
    .context("received a Modbus exception response")?;
```

and then the program might produce either a Modbus exception response:

```text
Error: received a Modbus exception response

Caused by:
    Gateway target device failed to respond
```

or some other I/O error (in this case simulated by connecting to a server that sends random bytes instead of Modbus frames back):

```text
Error: communication error

Caused by:
    Connection reset by peer (os error 104)
```

Thanks for getting this working!

PR #218, opened 2023-08-25
Publicly export `ExceptionResponse` and `Exception` types

Stupremee:
By making these types public, one is able to get the specific modbus error from a Modbus server.

So this is a short-term solution for #169, but it's definitely not a good one.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -38,7 +38,9 @@ pub use self::slave::{Slave, SlaveId};
 mod codec;
 
 mod frame;
-pub use self::frame::{Address, FunctionCode, Quantity, Request, Response};
+pub use self::frame::{
+    Address, Exception, ExceptionResponse, FunctionCode, Quantity, Request, Response,
+};
 
 mod service;
 


uklotzde:
How about a code example? Could we add error handling to one of the existing examples or (even better) a doctest example?

Stupremee:
Yep that's a good idea. Will add it tomorrow!

src/lib.rs
```diff
@@ -38,7 +38,9 @@ pub use self::slave::{Slave, SlaveId};
 mod codec;
 
 mod frame;
-pub use self::frame::{Address, FunctionCode, Quantity, Request, Response};
+pub use self::frame::{
+    Address, Exception, ExceptionResponse, FunctionCode, Quantity, Request, Response,
```
benjamin-nw, line 42: I think its a bad idea to make `ExceptionResponse` public with its fields also public.

We should add at least 2 methods for the `ExceptionResponse`:
- `exception()` that returns the `Exception`
- `function_code()` that returns the `FunctionCode`

And we should also enable a simple creation of the `ExceptionResponse` with a `new` method, or a method on `Request` / `Response` maybe to create a `ExceptionResponse` from it.

benjamin-nw: After thinking about it more, I think only `Exception` should be in the public API.

The user shouldn't have to know the existence of `ExceptionResponse`, the server should be able to build the appropriate `Response` type (either `Response` or `ExceptionResponse`) by the return of the `call` function.

See #237 for more context.

uklotzde:
Closing in favor of #246.

Issue #98, opened 2021-08-30
Any example for ExceptionResponse?


ArcherGu:
I'm a rust newcomer and am building a modbus server, and when the client sends the wrong address I need to give feedback on the error message, whether there are any examples of ExpressinResponse

justinyhuang:
Any response to this question? it appears to me that the exception definitions are in `frame` which is private.
How could a server implementation produce an appropriate exception response in this case?

glennpierce:
I wanted to match on 

Custom { kind: Other, error: ExceptionResponse { function: 3, exception: GatewayTargetDevice } } => {

}

but you are right frame is private where I find myself needing a few things from it.
In the end I converted the error to a string and did a match (yuck)

LenRemmerswaal:
I would think that you could have the modbus server return an error response, by returning  from call() like so:
```
 . . . 
    let fncode = req_to_fn_code(&req);
    return std::future::ready(Err(ExceptionResponse {
        function: fncode,
        exception: Exception::IllegalFunction,   // Or any other kind of Modbus exception
    }));
 . . .
```
Like mentioned before, for this to work, the following needs to be published:
- frame::ExceptionResponse
- frame::Exception
- codec::req_to_fn_code()

The first 2 could be included in the prelude, like Request and Response are.
The last 1 I am not sure about.

If you return with another kind of error than an ExceptionResponse, like std::io::Error, the application compiles, but the connection is forcibly closed by the server when the error is returned.

PR #226, opened 2023-10-17
More possibilities with Modbus-Exceptions

### LINKED ISSUE #169: Client: how to identify Modbus exception code sent by server in response

emcell:
This got a little bigger than I initially expected. I hope you're ok with my changes.

### What I've done so far
- Made some types of `frame` module public
- Converted `FunctionCode` to an enum and replaced it in a lot of places in codec module. I've done this because I needed the function code in the service handlers to send the Correct `ExceptionResponse` object. 
- Added the `GetFunctionCode` trait and implemented it on `Request` and `Response`. This is also necessary to generate the correct `ExceptionResponse` in the handlers.
- Added the `ExtractExceptionResponse` trait. This is for an easy way to downcast the `std::io::Error` in the clients to `ExceptionResponse`
- Added a new example that combines the following things:
  - async tcp-server
  - async rtu-server
  - An example of a Service that really uses async and not `futures::ready`
  - Specifying ExceptionResponses in the Service Layer
  - extracting ExceptionResponse from the client response

I started this because I didn't find ways to specify Modbus-Exception and didn't find a way to extract those in the clients.

I Think this PR should fix #169 and #98. 

### What could be better here?

I think It would be better to be able to Just return a `Result<Response, Exception>` from the Service and figure the correct function_code out after the service handler and automatically convert the `Exception` into `ExceptionResult`. 
But I didn't want to make an even bigger change without speaking to to you in the first place.

### DIFF
--- a/src/frame/mod.rs
+++ b/src/frame/mod.rs
@@ -12,7 +12,80 @@ use std::{borrow::Cow, error, fmt};
 use crate::bytes::Bytes;
 
 /// A Modbus function code is represented by an unsigned 8 bit integer.
-pub type FunctionCode = u8;
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum FunctionCode {
+    ReadCoils,                  //=> 0x01,
+    ReadDiscreteInputs,         // => 0x02,
+    WriteSingleCoil,            //(_, _) => 0x05,
+    WriteMultipleCoils,         //(_, _) => 0x0F,
+    ReadInputRegisters,         //(_) => 0x04,
+    ReadHoldingRegisters,       //(_) => 0x03,
+    WriteSingleRegister,        //(_, _) => 0x06,
+    WriteMultipleRegisters,     //(_, _) => 0x10,
+    MaskWriteRegister,          //(_, _, _) => 0x16,
+    ReadWriteMultipleRegisters, //(_) => 0x17,
+    Custom(u8),                 // => code,
+}
+
+impl Into<u8> for FunctionCode {
+    fn into(self) -> u8 {
+        match self {
+            FunctionCode::ReadCoils => 0x01,
+            FunctionCode::ReadDiscreteInputs => 0x02,
+            FunctionCode::WriteSingleCoil => 0x05,
+            FunctionCode::WriteMultipleCoils => 0x0f,
+            FunctionCode::ReadInputRegisters => 0x04,
+            FunctionCode::ReadHoldingRegisters => 0x03,
+            FunctionCode::WriteSingleRegister => 0x06,
+            FunctionCode::WriteMultipleRegisters => 0x10,
+            FunctionCode::MaskWriteRegister => 0x16,
+            FunctionCode::ReadWriteMultipleRegisters => 0x17,
+            FunctionCode::Custom(code) => code,
+        }
+    }
+}
+
+impl From<u8> for FunctionCode {
+    fn from(value: u8) -> Self {
+        match value {
+            0x01 => FunctionCode::ReadCoils,
+            0x02 => FunctionCode::ReadDiscreteInputs,
+            0x05 => FunctionCode::WriteSingleCoil,
+            0x0f => FunctionCode::WriteMultipleCoils,
+            0x04 => FunctionCode::ReadInputRegisters,
+            0x03 => FunctionCode::ReadHoldingRegisters,
+            0x06 => FunctionCode::WriteSingleRegister,
+            0x10 => FunctionCode::WriteMultipleRegisters,
+            0x16 => FunctionCode::MaskWriteRegister,
+            0x17 => FunctionCode::ReadWriteMultipleRegisters,
+            value => FunctionCode::Custom(value),
+        }
+    }
+}
+
+impl std::fmt::Display for FunctionCode {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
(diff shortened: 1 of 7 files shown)


src/frame/mod.rs
```diff
@@ -12,7 +12,80 @@ use std::{borrow::Cow, error, fmt};
 use crate::bytes::Bytes;
 
 /// A Modbus function code is represented by an unsigned 8 bit integer.
-pub type FunctionCode = u8;
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum FunctionCode {
+    ReadCoils,                  //=> 0x01,
```
uklotzde, line 17: These comments introduce unneeded redundancies and do not provide any value.

src/frame/mod.rs
```diff
@@ -12,7 +12,80 @@ use std::{borrow::Cow, error, fmt};
...
+    ReadInputRegisters,         //(_) => 0x04,
+    ReadHoldingRegisters,       //(_) => 0x03,
+    WriteSingleRegister,        //(_, _) => 0x06,
+    WriteMultipleRegisters,     //(_, _) => 0x10,
+    MaskWriteRegister,          //(_, _, _) => 0x16,
+    ReadWriteMultipleRegisters, //(_) => 0x17,
+    Custom(u8),                 // => code,
+}
+
+impl From<FunctionCode> for u8 {
```
uklotzde, line 30: The actual conversions should be implemented by a `const fn` method.

src/frame/mod.rs
```diff
@@ -12,7 +12,80 @@ use std::{borrow::Cow, error, fmt};
...
+            0x06 => FunctionCode::WriteSingleRegister,
+            0x10 => FunctionCode::WriteMultipleRegisters,
+            0x16 => FunctionCode::MaskWriteRegister,
+            0x17 => FunctionCode::ReadWriteMultipleRegisters,
+            value => FunctionCode::Custom(value),
+        }
+    }
+}
+
+impl std::fmt::Display for FunctionCode {
```
uklotzde, line 66: I don't think that this is a suitable implementation of std::fmt::Display. Its more a like an extended debug formatting and is not needed. Function codes should be displayed as their inner `u8` type.

src/frame/mod.rs
```diff
@@ -271,6 +392,30 @@ pub struct ExceptionResponse {
     pub exception: Exception,
 }
 
+/// Convenience trait for downcasting `std::io::Error` to `ExceptionResponse`
```
uklotzde, line 395: Why do we need to introduce a public trait? What is the use case?

src/frame/mod.rs
```diff
@@ -36,6 +109,11 @@ pub(crate) type Word = u16;
 /// Number of items to process.
 pub type Quantity = u16;
 
+/// A Trait that is implemented on types that have a function code
```
uklotzde, line 112: This trait is not needed. Could be implemented independently on types using `const fn` methods.

uklotzde (changes requested):

uklotzde:
Thanks for contributing!

IMHO this PR contains too many unrelated changes. The `FunctionCode` newtype should be introduced by a separate PR.

src/frame/mod.rs
emcell: I've removed the Function code enum

src/frame/mod.rs
emcell: done

src/frame/mod.rs
emcell: since the whole type is removed. It's display method also doesnt exist anymore

src/frame/mod.rs
emcell: I've added an example to make this more clear. 
Just an easy way to downcast the std::io::Error to the ExceptionResponse if it is one. 
I'm also not really with this solution. The better way would be to introduce an Error type for the clients. Could look like this

```rust
pub enum Error {
    ModbusException(ModbusException),
    IoError(std::io::Error)
}
```

This this change it's clear for the api users that a ModbusException exists and can be extracted. On the other hand that could break a lot of builds. 
What do you think?

src/frame/mod.rs
emcell: I've deleted the trait

emcell:
any chance this is getting merged soon?

uklotzde:
Please give this PR a concise title that later becomes the Git commit message.

We also need a changelog entry that summarizes the changes and consequences for users of this crate. Ideally a single sentence for every notable API change. As short as possible.

Everyone is invited to do a review. We are managing this crate in our spare time without any monetary compensation. Please understand that paid projects might be prioritized. Thank you.

uklotzde:
I will close this PR after seeing no progress lately and because merge conflicts are pending.

Please open a new PR whenever you have something ready. Thank you.

PR #235, opened 2023-12-21
Make frame::Exception public

jacobsvante:
Today one has to match on the to_string() representation, which is sub-optimal to say the least.


jacobsvante:
Never mind, I missed PRs #218 and #231.

Issue #165, opened 2023-03-09
TCP Server: How to return Modbus exception codes?


uklotzde:
See also: #163

benjamin-nw:
Hello, I'm trying to use `tokio-modbus` to create a Modbus Server implementation, and I've also hit this wall.

I took a look at the code and the issues/PR this morning.

I think #226 is a good start and I'll elaborate on that further.

But I think that the downcast is a short fix, and will only postpone the problem to more complex cases. And I also don't like to show more private types to the public API.

The public API should stay simple with `Request`, `Response` and `Error` and nothing else.

## 1. Modify `Response` (breaking change)

The first idea is to add a variant to `Response`:

```rust
pub enum Reponse {
    // [other variants]...
    Exception(FunctionCode, Exception),
}
``` 

Technically, an exception is still a "valid" response from the Modbus Server.

#### Pros

- Removes the burden of having a `Result<Response, ExceptionResponse>` in `ResponsePDU`.
- Relax the trait requirements  for `Response` in the `Service` trait.

#### Cons

- Need to refacto a lot of Client code to handle the new `Response` and return a correct `Error`.

### Conclusion

I think that while this solution is simple, it could lead to misinterpretation. This `Response` variant should be well documented and clearly state that its an Error response.

For me, this solution highlights another big issue.

The fact that all `Error`s in this crate are `std::io::Error`.

## 2. Changing `std::io::Error` to an internal `Error` type with more granularity (breaking change)

This is a big change, but imo, it should ease the use, simplify the API and provide better error handling without modifying the `Response` part.

The `Error` type should allow the user to know what kind of error happened:
- modbus exception
- `io::Error`
- more if needed...

The `Error` type should also define if an error is recoverable (i.e: modbus exception), or not (i.e: server disconnected or others I/O errors).

#### Pros

- Allow the user to match on the specific error types
- Reduce misuse of the library error type

#### Cons

- This is a breaking change, we need to update the `std::io::Error` into something like `crate::error::Error`.
- Need to review some parts in Client and Server to update the logic accordingly.

### Example

I'm taking the idea of @emcell [here](https://github.com/slowtec/tokio-modbus/pull/226#discussion_r1363229912).

Because I think that's the way to go, I've only made a few small changes.

```rust
pub enum Error {
    /// This is a error returned by the Modbus Server and relative to Modbus.
    Exception(ModbusException),

    /// This is an error in the Modbus Client side (can or can't be related to Modbus).
[...]

PR #236, opened 2023-12-27
Replace `FunctionCode` type alias with enum

benjamin-nw:
This PR try to improve the usability of the `FunctionCode` type by adding an enum with modbus public function code, as well a custom function code defined by the user.

The goal is to reduce code duplication and ease the use of the `FunctionCode` in order to avoid mistakes.

It also adds the ability to retrieve the `FunctionCode` associated with a `Request`/`Response`.

This will be useful in the futur when we try to implement #165 because we need to have access to the type of `Request` we are handling in case of an error in the `server`.

I took inspiration from the work in #226 .

This is a Breaking Change as we are updating a Public type.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### Breaking Changes
 
-- `FunctionCode`: Change `type alias` to an `enum`.
+- `FunctionCode`: Replace `type` alias with `enum`.
 
 ## v0.9.0 (2023-07-26)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@
 ## v0.10.0 (Unreleased)
 
 - Exclude `rtu-server`/`tcp-server` from default features.
+- Feature: Retrieve the `FunctionCode` of a `Request`/`Response`.
+
+### Breaking Changes
+
+- `FunctionCode`: Change `type alias` to an `enum`.
 
 ## v0.9.0 (2023-07-26)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 # Changelog
 
+## v0.10.0 (Unreleased)
+
+- Exclude `rtu-server`/`tcp-server` from default features.
+
 ## v0.9.0 (2023-07-26)
 
 - Optimization: Avoid allocations when writing multiple coils/registers.
--- a/src/frame/mod.rs
+++ b/src/frame/mod.rs
@@ -86,18 +86,6 @@ impl FunctionCode {
     }
 }
 
-impl From<FunctionCode> for u8 {
-    fn from(fc: FunctionCode) -> Self {
-        fc.value()
-    }
-}
-
-impl From<u8> for FunctionCode {
-    fn from(value: u8) -> Self {
-        Self::new(value)
-    }
-}
-
 impl Display for FunctionCode {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "{}", self.value())
--- a/src/frame/mod.rs
+++ b/src/frame/mod.rs
@@ -7,12 +7,102 @@ pub(crate) mod rtu;
 #[cfg(feature = "tcp")]
 pub(crate) mod tcp;
 
-use std::{borrow::Cow, error, fmt};
+use std::{
+    borrow::Cow,
+    error,
+    fmt::{self, Display},
+};
 
 use crate::bytes::Bytes;
 
-/// A Modbus function code is represented by an unsigned 8 bit integer.
-pub type FunctionCode = u8;
+/// A Modbus function code.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum FunctionCode {
+    /// Modbus Function Code: `01` (`0x01`).
+    ReadCoils,
+    /// Modbus Function Code: `02` (`0x02`).
+    ReadDiscreteInputs,
+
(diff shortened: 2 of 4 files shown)


src/frame/mod.rs
```diff
@@ -7,12 +7,102 @@ pub(crate) mod rtu;
...
+            FunctionCode::WriteMultipleCoils => 0x0F,
+            FunctionCode::WriteMultipleRegisters => 0x10,
+            FunctionCode::MaskWriteRegister => 0x16,
+            FunctionCode::ReadWriteMultipleRegisters => 0x17,
+            FunctionCode::Custom(code) => code,
+        }
+    }
+}
+
+impl From<FunctionCode> for u8 {
```
flosse, line 89: I don't know if we really want this implicit conversion.

src/frame/mod.rs
```diff
@@ -7,12 +7,102 @@ pub(crate) mod rtu;
...
+    fn from(fc: FunctionCode) -> Self {
+        fc.value()
+    }
+}
+
+impl From<u8> for FunctionCode {
```
flosse, line 95: ... same here.

src/frame/mod.rs
uklotzde: FunctionCode -> u8: Ok. But what is the public use case?
u8 -> FunctionCode: Not recommended!

benjamin-nw: > FunctionCode -> u8: Ok. But what is the public use case?

None I think, the user only needs to use the enum variant. Thus, I could remove them both.

benjamin-nw: Should the `new` and `value` functions be public too then ?

src/frame/mod.rs
```diff
@@ -7,12 +7,90 @@ pub(crate) mod rtu;
...
+            FunctionCode::MaskWriteRegister => 0x16,
+            FunctionCode::ReadWriteMultipleRegisters => 0x17,
+            FunctionCode::Custom(code) => code,
+        }
+    }
+}
+
+impl Display for FunctionCode {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.value())
```
uklotzde, line 91: Directly and more idiomatic:

```suggestion
        self.value().fmt(f)
```

src/frame/mod.rs
```diff
@@ -138,6 +216,34 @@ impl<'a> Request<'a> {
             Disconnect => Disconnect,
         }
     }
+
+    /// Get the [`FunctionCode`] of the [`Request`].
+    #[must_use]
+    pub const fn function_code(&self) -> FunctionCode {
+        use Request::*;
```
uklotzde, line 223:
```suggestion
        use Self::*;
```

Could also be replaced elsewhere for both `Request` and `Response`.

src/frame/mod.rs
```diff
@@ -95,7 +173,7 @@ pub enum Request<'a> {
     /// A raw Modbus request.
     /// The first parameter is the Modbus function code.
     /// The second parameter is the raw bytes of the request.
-    Custom(FunctionCode, Cow<'a, [u8]>),
+    Custom(u8, Cow<'a, [u8]>),
```
uklotzde, line 176: Why do we need to use the plain `u8` instead of the newtype wrapper here?

src/frame/mod.rs
benjamin-nw: This is not supported by the compiler. Or I'm doing something wrong 🤔 

```shell
error[E0432]: unresolved import `Self`
   --> src/frame/mod.rs:338:13
    |
338 |         use Self::*;
    |             ^^^^ `Self` cannot be used in imports
```

src/frame/mod.rs
uklotzde: Thanks for the clarification. I didn't consider the module scopes.

uklotzde:
The only thing that's missing now is a changelog entry.

benjamin-nw:
How do you want to handle the release of the `0.10` version ?

Do you push everything on `main` until the new version is released ? Or do you have a branch for a new version ?

Can I add the `## 0.10.0 (Unreleased)` entry in the Changelog here ?

My issue is, will other PR for 0.10 be able to access this changes on `main` or another branch ?

uklotzde:
Please cherry-pick the commit from #239 and add an entry for this PR to avoid merge conflicts. The commits will be squashed when we merge this PR afterwards.

uklotzde:
We usually only do _trunk based development_. If someone needs to maintain a legacy version for an extended period of time they could fork the repo.

benjamin-nw:
First time doing a cherry-pick.

I've first rebased on main and then cherry-picked your commit.

Did I do it right ?

uklotzde:
After

> First time doing a cherry-pick.
> 
> I've first rebased on main and then cherry-picked your commit.
> 
> Did I do it right ?

Never rebase, only merge main. Otherwise all review comments become dangling.

CHANGELOG.md
```diff
@@ -3,6 +3,15 @@
...
 # Changelog
 
+## v0.10.0 (Unreleased)
+
+- Exclude `rtu-server`/`tcp-server` from default features.
+- Feature: Retrieve the `FunctionCode` of a `Request`/`Response`.
+
+### Breaking Changes
+
+- `FunctionCode`: Change `type alias` to an `enum`.
```
uklotzde, line 13:
```suggestion
- `FunctionCode`: Change `type` alias to an `enum`.
```

uklotzde (approved):

PR #239, opened 2024-01-02
Exclude "rtu-server"/"tcp-server" from default features

uklotzde:
Enabling those _experimental_ features by default was never intended.


uklotzde:
The changelog entry should be moved to _Breaking Changes_ after #236 has been merged to avoid merge conflicts.

PR #244, opened 2024-01-04
server: Remove unused trait bounds from `Service::call()` future result

uklotzde:


uklotzde:
Since this trait is only supposed to be called from one of the server skeletons I don't consider this removal a breaking change. Otherwise it would be one.

flosse (approved):

PR #245, opened 2024-01-04
server: Add DynamicService trait and adapter

uklotzde:


uklotzde:
- PoC
- Based on #244

uklotzde:
Not helpful without a concrete use case and example. Maybe later.

PR #248, opened 2024-02-02
refacto: new `Client` API

creberust:
The new `Client` API enables the user to have 2 errors returned when calling a Modbus function.

1. `std::io::Error`: An error occured on the network side.
2. `tokio_modbus::Exception`: An error occured on the Modbus server.

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -6,7 +6,7 @@
 use std::{
     borrow::Cow,
     fmt::Debug,
-    io::{Error, ErrorKind},
+    io::{self, Error, ErrorKind},
 };
 
 use async_trait::async_trait;
@@ -26,24 +26,39 @@ pub mod sync;
 #[async_trait]
 pub trait Client: SlaveContext + Send + Debug {
     /// Invoke a Modbus function
-    async fn call(&mut self, request: Request<'_>) -> Result<Response, Error>;
+    async fn call(&mut self, request: Request<'_>) -> io::Result<crate::Result<Response>>;
 }
 
 /// Asynchronous Modbus reader
 #[async_trait]
 pub trait Reader: Client {
     /// Read multiple coils (0x01)
-    async fn read_coils(&mut self, _: Address, _: Quantity) -> Result<Vec<Coil>, Error>;
+    async fn read_coils(
+        &mut self,
+        addr: Address,
+        cnt: Quantity,
+    ) -> io::Result<crate::Result<Vec<Coil>>>;
 
     /// Read multiple discrete inputs (0x02)
-    async fn read_discrete_inputs(&mut self, _: Address, _: Quantity) -> Result<Vec<Coil>, Error>;
+    async fn read_discrete_inputs(
+        &mut self,
+        addr: Address,
+        cnt: Quantity,
+    ) -> io::Result<crate::Result<Vec<Coil>>>;
 
     /// Read multiple holding registers (0x03)
-    async fn read_holding_registers(&mut self, _: Address, _: Quantity)
-        -> Result<Vec<Word>, Error>;
+    async fn read_holding_registers(
+        &mut self,
+        addr: Address,
+        cnt: Quantity,
+    ) -> io::Result<crate::Result<Vec<Word>>>;
 
     /// Read multiple input registers (0x04)
-    async fn read_input_registers(&mut self, _: Address, _: Quantity) -> Result<Vec<Word>, Error>;
+    async fn read_input_registers(
+        &mut self,
+        addr: Address,
+        cnt: Quantity,
+    ) -> io::Result<crate::Result<Vec<Word>>>;
 
     /// Read and write multiple holding registers (0x17)
     ///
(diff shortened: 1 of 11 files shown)


src/service/tcp.rs
```diff
@@ -3,7 +3,7 @@
 
 use std::{
     fmt,
-    io::{Error, ErrorKind},
+    io::{self, Error, ErrorKind},
```
uklotzde, line 6: Only importing `io` and using `io::Error` and `io::ErrorKind` explicitly would be more consistent and less confusing.

uklotzde (commented):
Thanks for implementing the nested Result proposal. Looks good so far. It nicely separates technical from semantic/logic errors.

creberust:
Hi,

Thanks for the feedback !

I need to also update the examples with the new client API to correctly `assert` on error.

I'll add the `CHANGELOG` entry when I'm ready with the PR 👍🏻 .

uklotzde:
> I need to also update the examples with the new client API to correctly `assert` on error.

Updating the examples will hopefully also reveal if this approach is usable in practice.

src/client/sync/mod.rs
```diff
@@ -160,31 +193,43 @@ impl Reader for Context {
...
             self.async_ctx.write_single_register(addr, data),
         )
     }
 
-    fn write_multiple_registers(&mut self, addr: Address, data: &[Word]) -> Result<()> {
+    fn write_multiple_registers(
+        &mut self,
+        addr: Address,
+        data: &[Word],
+    ) -> io::Result<crate::Result<()>> {
```
uklotzde, line 212: I wonder if it would make sense to introduce a type alias for `io::Result<crate::Result<T>>`.

Rename `crate::Result` into `ModbusResult` and use `pub type Result<T> = io::Result<ModbusResult<T>>` as nested, crate-level `Result` type?

creberust: We could add 2 type aliases in `lib.rs`:
- `pub type ModbusResult<T> = Result<T, Exception>;`
- `pub type Result<T> = io::Result<ModbusResult<T>>;`

This will reduce the boilerplate of having to write `io::Result<crate::Result<()>>` everytime.

I'll update it to see if it's easier to understand this way, if not, we can revert the commit.

creberust: I think we even don't need `ModbusResult` at all.

Because it's only used on the `server` side and it's a simple `Result<Response, Exception>`.

I think we can only keep the type alias `Result<T> = io::Result<result::Result<T, Exception>>`

uklotzde: Minimizing the publicly visible and accessible parts of an API is recommended.

The `crate Result` alias is an exception to this rule, because it appears so often in the public (client) API. The client API should be the main focus.

creberust:
I tried it with our `modbus` client wrapper around `tokio_modbus` that does automatic re-connection on `io::Error`, and it works great ! 🎉 I can match on `io::Error` and let pass the modbus `Response` with the data or an exception.

I need to try with a more complex use case with multiple functions called asynchronously and in parallel. But things are looking great already !

creberust:
I have just one small question regarding the use of `unreachable` in the client side.

Can you confirm that the paths with `unreachable!` are indeed not supposed to happen ? And if it's reached, it's a bug in the client codec ?

uklotzde:
> I have just one small question regarding the use of `unreachable` in the client side.
> 
> Can you confirm that the paths with `unreachable!` are indeed not supposed to happen ? And if it's reached, it's a bug in the client codec ?

Paths marked as `unreachable!()` must never executed. Otherwise it is a bug. It must not be used for handling unexpected data that might occur, even if the spec disallows it.

Sometimes you simply cannot proof statically that certain paths are never executed. In this case `unreachable!()` is used for documentation purposes.

creberust:
Agreed,

For me, going down the `unreachable` path in this case:
```rust
        self.client
            .call(Request::ReadCoils(addr, cnt))
            .await
            .map(|modbus_rsp| {
                modbus_rsp.map(|rsp| match rsp {
                    Response::ReadCoils(mut coils) => {
                        debug_assert!(coils.len() >= cnt.into());
                        coils.truncate(cnt.into());
                        coils
                    }
                    others => {
                        unreachable!(
                            "unexpected response code: {} (request code: {})",
                            others.function_code(),
                            FunctionCode::ReadCoils
                        )
                    }
                })
            })
```

Is a bug in the `tokio_modbus` lib and _should_ never happen.

Are we ok with that ? If so, I think the PR is ready 👍🏻

uklotzde:
> For me, going down the `unreachable` path in this case:

Could you add a comment that points the reader where this is handled? Because at first sight it looks like the code would panic on invalid inputs.

creberust:
Moved `verify_response_header` function in the parent module to use the same function in `rtu` and `tcp` impl.

Added dev doc comments to explain why the `unreachable!` macro is used here.

Improved error message in `unreachable!` to ease the bug report handling.

Is it ok to have a link directly to the github issue page in the error message ?

src/client/mod.rs
```diff
@@ -140,7 +140,8 @@ impl Reader for Context {
                         // NOTE: A call to `Client::call` implementation *MUST* always return the `Response` variant matching the `Request` one.
                         // TIPS: This can be ensured via a call to `verify_response_header`( in 'src/service/mod.rs') before returning from `Client::call`.
                         unreachable!(
-                            "unexpected response code: {} (request code: {})",
+                            "unexpected response code: {} (request code: {})\n\
+                            note: Please fill an issue at `https://github.com/slowtec/tokio-modbus/issues` with a minimal example reproducing this bug.",
```
uklotzde, line 144: "fill" -> "file"?

This verbose message is duplicated many times. Could this be avoided by externalizing the composition of the error message into a function? Does not need to be efficient, because it should never be called.

uklotzde (changes requested):

uklotzde:
Sorry for the delay. I hope to review your PR again soon. Please be patient :see_no_evil:

uklotzde:
Feedback from other users of this library would greatly appreciated. Those are the ones that are most affected by this substantial API refactoring.

uklotzde:
https://github.com/creberust/tokio-modbus/pull/1

CHANGELOG.md
```diff
@@ -3,6 +3,28 @@
...
+```rust
+pub type Result<T> = Result<Result<T, Exception>, std::io::Error>
+```
+
+- Example: Update all examples with new `Client` API.
+
+### Breaking Changes
+
+- Client: All the functions signatures in `Client`, `Reader` and `Writer` traits
+  have changed the return type.
```
uklotzde, line 20:
```suggestion
- Client: All methods in the `Client`, `Reader`, and `Writer` traits now return nested `Result` values that both need to be handled explicitly.
```

uklotzde (commented):
Some suggestions for the wording in the changelog.

CHANGELOG.md
```diff
@@ -3,6 +3,28 @@
...
+
+  ```diff
+  async fn read_coils(&mut self, _: Address, _: Quantity)
+  - -> Result<Vec<Coil>, std::io::Error>;
+  + -> Result<Result<Vec<Coil>, Exception>, std::io::Error>;
+  ```
+
```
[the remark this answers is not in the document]
uklotzde: Please add newlines and apply (re-)formatting as needed.

uklotzde (approved):

PR #253, opened 2024-03-26
Upgrade rustls-pemfile and tokio-rustls

gustavowd:


flosse:
@gustavowd thanks for contributing! Would you mind to run `cargo fmt` and also fix the tests?

gustavowd:
In order to compile the aws-lc-sys v0.13.3 library, that is request for the new version of rustls i have to install some packets in ubuntu. I don´t now how to do this in the autotest. This are the packets:

sudo apt install bindgen cmake libclang-dev

uklotzde (approved):
Thanks! Upgrading the `rustls` dependency could now be done in a separate step.

PR #254, opened 2024-03-26
Fix panic on mismatching request/response function codes

### LINKED ISSUE #252: tokio-modbus 0.12 entering unreachable code in any response error
Hi,

Since tokio-modbus 0.12 i'm facing several application crashs. The reason is that in any returning error code the tokio-modbus enters an unreachable code. I'm communicating with a commercial equipament, which sometimes returns a invalid response. In tokio-modbus 0.11 a handle this erros, but now it always causes a application crash.

Mar 26 08:12:14 raspberrypi gateway.sh[29153]: thread 'main' panicked at /home/gateway/.cargo/registry/src/index.crates.io-1cd66030c949c28d/tokio-modbus-0.12.0/src/client/mod.rs:230:25:
Mar 26 08:12:14 raspberrypi gateway.sh[29153]: internal error: entered unreachable code: unexpected response code: 4 (request code: 3)
Mar 26 08:12:14 raspberrypi gateway.sh[29153]: note: Please report the issue at `https://github.com/slowtec/tokio-modbus/issues` with a minimal example reproducing this bug.
Mar 26 08:12:14 raspberrypi gateway.sh[29153]: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

Thanks in advance,
Gustavo

uklotzde:
Fixes #252.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -40,18 +40,32 @@ pub mod slave;
 pub use self::slave::{Slave, SlaveId};
 
 mod codec;
+
 mod error;
+pub use self::error::ResponseError;
 
 mod frame;
 pub use self::frame::{Address, Exception, FunctionCode, Quantity, Request, Response};
 
-/// Specialized [`std::result::Result`] type for `Modbus` client API.
+/// Specialized [`std::result::Result`] type for type-checked responses of the _Modbus_ client API.
+///
+/// The payload is generic over the response type.
 ///
 /// This [`Result`] type contains 2 layers of errors.
 ///
 /// 1. [`std::io::Error`]: An error occurred while performing I/O operations.
-/// 2. [`Exception`]: An error occurred on the `Modbus` server.
-pub type Result<T> = std::io::Result<std::result::Result<T, Exception>>;
+/// 2. [`ResponseError`]: An error occurred on the _Modbus_ server or when matching the received response with the request.
+pub type Result<T> = std::io::Result<std::result::Result<T, ResponseError>>;
+
+/// Result of a generic [`Request`].
+///
+/// Contains a generic [`Response`] if successful or an [`Exception`] if the
+/// server responded with an error.
+///
+/// The function codes and other properties of the request and response
+/// are not matched against each other. Any received response is passed-through
+/// unchecked.
+pub type ResponseResult = std::io::Result<std::result::Result<Response, Exception>>;
 
 mod service;
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -45,7 +45,9 @@ mod error;
 pub use self::error::ResponseError;
 
 mod frame;
-pub use self::frame::{Address, Exception, FunctionCode, Quantity, Request, Response};
+pub use self::frame::{
+    Address, Exception, ExceptionResponse, FunctionCode, Quantity, Request, Response,
+};
 
 /// Specialized [`std::result::Result`] type for type-checked responses of the _Modbus_ client API.
 ///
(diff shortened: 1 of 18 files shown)


uklotzde:
The seconds commit introduces another error variant that replaces `std::io::Error` in case of mismatching headers between request and response. All error variants include the actual response result.

src/lib.rs
```diff
@@ -40,18 +40,27 @@ pub mod slave;
...
 /// This [`Result`] type contains 2 layers of errors.
 ///
 /// 1. [`std::io::Error`]: An error occurred while performing I/O operations.
-/// 2. [`Exception`]: An error occurred on the `Modbus` server.
-pub type Result<T> = std::io::Result<std::result::Result<T, Exception>>;
+/// 2. [`ResponseError`]: An error occurred on the _Modbus_ server or when matching the received response with the request.
+pub type Result<T> = std::io::Result<std::result::Result<T, ResponseError>>;
+
+/// Result of a generic [`Request`].
+pub type ResponseResult = Result<Response>;
```
flosse, line 63:
```suggestion
type ResponseResult = Result<Response>;
```

For me, it reads much better when standard types are used. An alias is introduced here that I as a user don't recognize as such so quickly.

uklotzde: https://github.com/slowtec/tokio-modbus/pull/254/commits/a12143943f7520470dfd2f93494672bafe958733

flosse:
I'm sorry, where exactly is the fix?

flosse (approved):

src/service/rtu.rs
```diff
@@ -61,10 +62,31 @@ where
...
-            ResponsePdu(Err(err)) => Ok(Err(err.exception)),
+        let ResponsePdu(result) = res_adu.pdu;
+
+        // Match headers of request and response.
+        if let Err(message) = verify_response_header(&req_hdr, &res_adu.hdr) {
+            return Ok(Err(ResponseError::MismatchingHeaders { message, result }));
+        }
+
+        // Match function codes of request and response.
+        let rsp_function_code = match &result {
```
uklotzde, line 73: The fix (RTU).

src/service/tcp.rs
```diff
@@ -84,10 +85,31 @@ where
...
+        if let Err(message) = verify_response_header(&req_hdr, &res_adu.hdr) {
+            return Ok(Err(ResponseError::MismatchingHeaders { message, result }));
         }
+
+        // Match function codes of request and response.
+        let rsp_function_code = match &result {
+            Ok(response) => response.function_code(),
+            Err(ExceptionResponse { function, .. }) => *function,
+        };
+        if req_function_code != rsp_function_code {
```
uklotzde, line 100: The fix (TCP).

uklotzde:
I am still unsure if it would be more suitable to replace the outer `std::io::Error` with a new error type or the inner type as done here. Where do protocol errors belong?

uklotzde:
The alternative approach of replacing the outer `std::io::error` with a custom error type and preserving `Exception` as the inner error type turned out to be much more elegant.

The different sources of errors are now separated more consistently, i.e. client/server communication vs. server-side processing.

flosse:
> I am still unsure if it would be more suitable to replace the outer `std::io::Error` with a new error type or the inner type as done here. Where do protocol errors belong?

In general I'd look on the layers.
```
Application -> Modbus -> Transport
```
So my error design would be

```rust
enum MyAppError {
  Modbus(ModbusError),
  // other app specific
}
enum ModbusError {
  Transport(TransportError).
  // other modbus specific
}
enum TransportError {
  // ...
}
```

So the top level error in this library would be a very specific error where `io::Error` is just one variant.

flosse:
> The alternative approach of replacing the outer `std::io::error` with a custom error type and preserving `Exception` as the inner error type turned out to be much more elegant.

:joy: it looks like you had the same idea :smile: 

:100:

uklotzde:
Keeping the `Exception` separate as a _not-really-an-error_ without wrapping it justifies the nested Result pattern.

Issue #264, opened 2024-07-16
Cant ignore Modbus RTU Request


philippderdiedas:
Commit b4819768309dffce04145984dafdeb2a6a5355d6 prevents modbus (rtu) server from ignoring response.
trait requires Response to be `crate::Response` instead of own type e.g. `Option<crate::Response>`
So #141 `Ok(None)` does not work anymore

uklotzde:
Looks like this feature got lost during the refactoring. It was informally proposed as a solution, but no one has provided a test for this use case.

uklotzde:
Please test #265.

You may also add regression tests for your use case if you want to make sure that it remains covered.

philippderdiedas:
works like a charm 👍

PR #267, opened 2024-07-21
Fix rtu-server-address example

uklotzde:
The example should've served as a regression test for #264. But the implementation was wrong.


flosse (approved):
thx!

PR #273, opened 2024-08-28
Remove obsolete FunctionCode::Disconnect

uklotzde:
Use AsyncWriteExt::shutdown() to release all resources.

### DIFF
--- a/src/frame/mod.rs
+++ b/src/frame/mod.rs
@@ -79,8 +79,6 @@ pub enum FunctionCode {
 
     /// Custom Modbus Function Code.
     Custom(u8),
-
-    Disconnect,
 }
 
 impl FunctionCode {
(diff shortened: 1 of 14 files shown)


src/frame/mod.rs
```diff
@@ -181,17 +174,6 @@ pub enum Request<'a> {
     /// The first parameter is the Modbus function code.
     /// The second parameter is the raw bytes of the request.
     Custom(u8, Cow<'a, [u8]>),
-
-    /// A poison pill for stopping the client service and to release
-    /// the underlying transport, e.g. for disconnecting from an
-    /// exclusively used serial port.
-    ///
-    /// This is an ugly workaround, because `tokio-proto` does not
```
uklotzde, line 189: Does no longer apply.

uklotzde:
Anyone? I only did a quick test with a USB serial adapter (RTU), basically runing the RTU example code in a loop.

Unfortunately, I don't remember the details what didn't work in the past with `tokio-proto`.

uklotzde:
We should probably merge this PR and wait for reports from users that are already using `main`. Needs some testing before a release.

flosse (approved):

uklotzde:
Let's give it a try.

Issue #123, opened 2022-09-01
Support for custom functions in RTU


vdominguez1993:
I have notices that custom functions are supported and there is even an example for TCP.

But I was testing in RTU and I found that the custom command is sent, but the response is never processed.
This is in part functions are not supported in:
```Rust fn get_response_pdu_len(adu_buf: &BytesMut) -> Result<Option<usize>>```

Additionally in this funcion some quick fix can be done for supporting exception code responses for all functions, changing the match arm from `0x81..=0xAB` to `0x81..=0xFF`. As all the exceptions share the same format regardless of the function code.

I guess the initial problem of not supporting this is how to determine the start and end of a frame.
I have tested some changes in the decode functions and I kind of found a way to deal with it. It's not the best option, but it's something.

```diff
diff --git a/src/codec/rtu.rs b/src/codec/rtu.rs
index 60a4cf8..5e39da0 100644
--- a/src/codec/rtu.rs
+++ b/src/codec/rtu.rs
@@ -11,6 +11,7 @@ use log::{debug, error, warn};
 use smallvec::SmallVec;
 use std::io::{Cursor, Error, ErrorKind, Result};
 use tokio_util::codec::{Decoder, Encoder};
+use std::time::{Instant, Duration};
 
 // [MODBUS over Serial Line Specification and Implementation Guide V1.02](http://modbus.org/docs/Modbus_over_serial_l
ine_V1_02.pdf), page 13
 // "The maximum size of a MODBUS RTU frame is 256 bytes."
@@ -105,6 +106,8 @@ pub(crate) struct RequestDecoder {
 #[derive(Debug, Default, Eq, PartialEq)]
 pub(crate) struct ResponseDecoder {
     frame_decoder: FrameDecoder,
+    last_time: Option<Instant>,
+    last_size: usize,
 }
 
 #[derive(Debug, Default, Eq, PartialEq)]
@@ -165,8 +168,15 @@ fn get_response_pdu_len(adu_buf: &BytesMut) -> Result<Option<usize>> {
                     // incomplete frame
                     return Ok(None);
                 }
-            }
-            0x81..=0xAB => 2,
+            },
+            // For custom functions, use all the buffer
+            65..=75 | 100..=110 => if adu_buf.len() > 3 {
+                adu_buf.len() - 3
+            } else {
+                return Ok(None);
+            },
+            // Above this value it means an exception code as the first bit is 1
+            0x81..=0xFF => 2,
             _ => {
[...]

PR #274, opened 2024-09-09
Implement `Report Server ID` (function code 17)

snejugal:
This MR implements the `Report Server ID` function (code 17) from the spec.

I tried to work around the absence of this function using `Custom` requests and responses, but it turns out `tokio-modbus` does not support them properly. Since this function is defined in the spec, I decided to add native support for this function.

### DIFF
--- a/src/codec/mod.rs
+++ b/src/codec/mod.rs
@@ -81,6 +81,7 @@ impl<'a> TryFrom<Request<'a>> for Bytes {
                     data.put_u16(*w);
                 }
             }
+            ReportServerId => {}
             MaskWriteRegister(address, and_mask, or_mask) => {
                 data.put_u16(address);
                 data.put_u16(and_mask);
@@ -145,6 +146,14 @@ impl From<Response> for Bytes {
             WriteMultipleCoils(address, quantity) | WriteMultipleRegisters(address, quantity) => {
                 data.put_u16(address);
                 data.put_u16(quantity);
+            }
+            ReportServerId(server_id, run_indication, additional_data) => {
+                data.put_u8(2 + u8_len(additional_data.len()));
+                data.put_u8(server_id);
+                data.put_u8(if run_indication { 0xFF } else { 0x00 });
+                for b in additional_data {
+                    data.put_u8(b);
+                }
             }
             WriteSingleRegister(address, word) => {
                 data.put_u16(address);
@@ -224,6 +233,7 @@ impl TryFrom<Bytes> for Request<'static> {
                 }
                 WriteMultipleRegisters(address, data.into())
             }
+            0x11 => ReportServerId,
             0x16 => {
                 let address = rdr.read_u16::<BigEndian>()?;
                 let and_mask = rdr.read_u16::<BigEndian>()?;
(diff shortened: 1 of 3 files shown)


uklotzde:
Please ignore the pre-commit failure. I take care of that separately.

src/frame/mod.rs
```diff
@@ -322,6 +333,12 @@ pub enum Response {
     /// The second parameter contains the amount of register that have been written
     WriteMultipleRegisters(Address, Quantity),
 
+    /// Response to a `ReportServerId` request
+    /// The first parameter contains the server ID
+    /// The second parameter indicates whether the servier is running
```
uklotzde, line 338:
```suggestion
    /// The second parameter indicates whether the server is running
```

src/codec/mod.rs
```diff
@@ -146,6 +147,14 @@ impl From<Response> for Bytes {
                 data.put_u16(address);
                 data.put_u16(quantity);
             }
+            ReportServerId(server_id, run_indication, additional_data) => {
+                data.put_u8(2 + u8_len(additional_data.len()));
```
uklotzde, line 151: - What is the maximum length of `additional_data` according to the spec?
- What happens if you pass more data than allowed? The library should reject such a request instead of panicking.

uklotzde: - Is the server allowed to respond with empty `additional` data?

Ideally, there would be tests for all those use cases.

uklotzde:
Merging `main` will fix the pre-commit failure.

src/codec/mod.rs
uklotzde: I just noticed that there don't seem to be range checks for the other requests :sweat_smile: So I guess we could leave it as is for now.

src/frame/mod.rs
snejugal: Fixed the typo, thanks!

snejugal:
Rebased on `main`, the precommit check should pass now.

As of support for custom functions, I think that a trait-based solution might be useful, though that will be a massive overhaul of the crate. Nonetheless, I will try to describe this idea in #123 so it can get some feedback before being implemented (by me or by someone else).

uklotzde:
In most cases you don't need to rebase. This would invalidate previous review comment. Each PR will be squashed into a single commit when merged.

src/frame/mod.rs
```diff
@@ -164,6 +169,9 @@ pub enum Request<'a> {
     /// The second parameter is the vector of values to write to the registers.
     WriteMultipleRegisters(Address, Cow<'a, [Word]>),
 
+    /// A request to report server ID.
```
uklotzde, line 172: This function code is marked as _Serial Line only_ in the spec. Should we add a comment?

snejugal: Indeed, added the note.

uklotzde:
> As of support for custom functions, I think that a trait-based solution might be useful, though that will be a massive overhaul of the crate.

Both the public API and the internal implementation could be improved and optimized in many ways.

uklotzde:
Technically this is a breaking change because `FunctionCode` is not marked as `non_exhaustive` (which should probably be added). But I tend to release it as v0.14.1.

Opinions?

uklotzde:
The breaking change only affects the (experimental) server-side API. Another reason why this should not be bundled in a single crate together with the client-side API.

PR #279, opened 2024-09-11
feat: add exception other

### LINKED ISSUE #278: Cannot obtain custom Exception code
Whilst trying to communicate with a Huawei inverter, read commands work as expected but if the session is not authorized (a proprietary thing), write commands fail with an exception code of `0x80` (128).

Upon inspecting the Transport error, this is what I get:
```
Transport(
    Custom {
        kind: InvalidData,
        error: "Invalid exception code",
    },
)
```
Source:
https://github.com/slowtec/tokio-modbus/blob/903951a0bbef3e459fe012a59ef58cc80a9d7963/src/codec/mod.rs#L412

And the [tokio_modbus::Exception](https://docs.rs/tokio-modbus/latest/tokio_modbus/enum.Exception.html) has only the following exception codes (which all are standard as far as I understand): 
```rust
#[repr(u8)]
pub enum Exception {
    IllegalFunction = 1,
    IllegalDataAddress = 2,
    IllegalDataValue = 3,
    ServerDeviceFailure = 4,
    Acknowledge = 5,
    ServerDeviceBusy = 6,
    MemoryParityError = 8,
    GatewayPathUnavailable = 10,
    GatewayTargetDevice = 11,
}
```
(I don't think something like `Other(u8)` would work here because of the `repr(u8)`, but maybe implementing `Into<u8>` instead of it would)

So, my problem is: How can I properly get the exception code? Are there any other ways to get it? Have I misunderstood something?

Thanks.

CosminPerRam:
Would close #278

This PR is just for a better insight on my problem and suggested changes, as such, it has breaking changes.

### DIFF
--- a/src/codec/mod.rs
+++ b/src/codec/mod.rs
@@ -385,7 +385,7 @@ impl TryFrom<Bytes> for ExceptionResponse {
             ));
         }
         let function = fn_err_code - 0x80;
-        let exception = Exception::try_from(rdr.read_u8()?)?;
+        let exception = Exception::from(rdr.read_u8()?);
         Ok(ExceptionResponse {
             function: FunctionCode::new(function),
             exception,
@@ -393,12 +393,10 @@ impl TryFrom<Bytes> for ExceptionResponse {
     }
 }
 
-impl TryFrom<u8> for Exception {
-    type Error = Error;
-
-    fn try_from(code: u8) -> Result<Self, Self::Error> {
+impl From<u8> for Exception {
+    fn from(code: u8) -> Self {
         use crate::frame::Exception::*;
-        let ex = match code {
+        match code {
             0x01 => IllegalFunction,
             0x02 => IllegalDataAddress,
             0x03 => IllegalDataValue,
@@ -408,11 +406,8 @@ impl TryFrom<u8> for Exception {
             0x08 => MemoryParityError,
             0x0A => GatewayPathUnavailable,
             0x0B => GatewayTargetDevice,
-            _ => {
-                return Err(Error::new(ErrorKind::InvalidData, "Invalid exception code"));
-            }
-        };
-        Ok(ex)
+            other => Other(other),
+        }
     }
 }
 
--- a/src/frame/mod.rs
+++ b/src/frame/mod.rs
@@ -387,22 +387,34 @@ impl Response {
 
 /// A server (slave) exception.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-#[repr(u8)]
 pub enum Exception {
-    IllegalFunction = 0x01,
-    IllegalDataAddress = 0x02,
-    IllegalDataValue = 0x03,
-    ServerDeviceFailure = 0x04,
-    Acknowledge = 0x05,
-    ServerDeviceBusy = 0x06,
-    MemoryParityError = 0x08,
-    GatewayPathUnavailable = 0x0A,
-    GatewayTargetDevice = 0x0B,
+    IllegalFunction, // 0x01
+    IllegalDataAddress, // 0x02
+    IllegalDataValue, // 0x03
+    ServerDeviceFailure, // 0x04
+    Acknowledge, // 0x05
+    ServerDeviceBusy, // 0x06
+    MemoryParityError, // 0x08
+    GatewayPathUnavailable, // 0x0A
+    GatewayTargetDevice, // 0x0B
+    Other(u8),
 }
 
 impl From<Exception> for u8 {
     fn from(from: Exception) -> Self {
-        from as u8
+        use crate::frame::Exception::*;
+        match from {
+            IllegalFunction => 0x01,
+            IllegalDataAddress => 0x02,
+            IllegalDataValue => 0x03,
+            ServerDeviceFailure => 0x04,
+            Acknowledge => 0x05,
+            ServerDeviceBusy => 0x06,
+            MemoryParityError => 0x08,
+            GatewayPathUnavailable => 0x0A,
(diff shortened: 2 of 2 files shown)


uklotzde:
Thanks for contributing!

Some general remarks for discussion.

I have noticed that the naming of the types `FunctionCode` and `Exception` is inconsistent. Shouldn't `Exception` better be renamed to `ExceptionCode` to reflect this duality?

The different naming of the fallback variants `Custom` and `Other` respectively is inconsistent. That's what I initially stumbled upon.

CosminPerRam:
> The different naming of the fallback variants Custom and Other respectively is inconsistent and confusing.

> Shouldn't Exception better be renamed to ExceptionCode to reflect this duality?

Yes, having consistency would be great to see, should this be done in this PR or open another one?

> Not sure which term is more common in this context.

I'm not sure either, but personally I'd go with `Custom`, sounds nicer in this context, as in 'one might have **custom** implementations' or 'this device returns a **custom** code'.

uklotzde:
We should address the naming of `Custom`/`Other` in this PR and defer the renaming of the type(s) to a follow-up PR.

Let's use `Custom` for now to avoid changing existing code.

src/frame/mod.rs
```diff
@@ -387,22 +387,34 @@ impl Response {
...
+    IllegalFunction,        // 0x01
+    IllegalDataAddress,     // 0x02
+    IllegalDataValue,       // 0x03
+    ServerDeviceFailure,    // 0x04
+    Acknowledge,            // 0x05
+    ServerDeviceBusy,       // 0x06
+    MemoryParityError,      // 0x08
+    GatewayPathUnavailable, // 0x0A
+    GatewayTargetDevice,    // 0x0B
+    Other(u8),
```
uklotzde, line 400: As proposed.

`/// None of the above.`

src/frame/mod.rs
```diff
@@ -420,6 +444,7 @@ impl Exception {
             MemoryParityError => "Memory parity error",
             GatewayPathUnavailable => "Gateway path unavailable",
             GatewayTargetDevice => "Gateway target device failed to respond",
+            Custom(_) => "Other",
```
uklotzde, line 447:
```suggestion
            Custom(_) => "Custom",
```

uklotzde:
```suggestion
            Custom(_) => "Custom",
```

uklotzde: We don't know if `Custom` actually contains a value that equals one of the above, i.e. _other_. Let's simply return the name of the variant.

src/codec/mod.rs
```diff
@@ -408,11 +406,8 @@ impl TryFrom<u8> for Exception {
             0x08 => MemoryParityError,
             0x0A => GatewayPathUnavailable,
             0x0B => GatewayTargetDevice,
-            _ => {
-                return Err(Error::new(ErrorKind::InvalidData, "Invalid exception code"));
-            }
-        };
-        Ok(ex)
+            custom => Custom(custom),
```
uklotzde, line 409:
```suggestion
            other => Custom(other),
```

uklotzde: Sorry for the confusion. The naming `other` in the match statement was semantically correct here.

CosminPerRam: Nice one, for sure.

src/frame/mod.rs
```diff
@@ -387,22 +387,46 @@ impl Response {
...
+    ServerDeviceBusy,
+    /// 0x08
+    MemoryParityError,
+    /// 0x0A
+    GatewayPathUnavailable,
+    /// 0x0B
+    GatewayTargetDevice,
+    /// None of the above.
+    ///
+    /// Although encoding one of the predefined values as this is possible, it is not recommended.
```
uklotzde, line 411: We could then add something like the following hint:

```
Instead of instantiating this variant directly prefer to use [`Self::new()`] to prevent such ambiguities.
```

src/codec/mod.rs
```diff
@@ -385,34 +385,17 @@ impl TryFrom<Bytes> for ExceptionResponse {
             ));
         }
         let function = fn_err_code - 0x80;
-        let exception = Exception::try_from(rdr.read_u8()?)?;
+        let exception = Exception::from(rdr.read_u8()?);
```
uklotzde, line 388:
```suggestion
        let exception = Exception::new(rdr.read_u8()?);
```

src/codec/mod.rs
```diff
@@ -385,34 +385,17 @@ impl TryFrom<Bytes> for ExceptionResponse {
...
-            0x06 => ServerDeviceBusy,
-            0x08 => MemoryParityError,
-            0x0A => GatewayPathUnavailable,
-            0x0B => GatewayTargetDevice,
-            _ => {
-                return Err(Error::new(ErrorKind::InvalidData, "Invalid exception code"));
-            }
-        };
-        Ok(ex)
+impl From<u8> for Exception {
```
uklotzde, line 396: I suggest to delete this `impl From`.

As a user of this crate you probably do not want to convert arbitrary `u8` values implicitly into a Modbus exception code with `.into()`. With `Exception::new()` we now have an explicit constructor to make such a conversion explicit.

I generally try to avoid `From`/`TryFrom` implementations between types of different abstraction levels. These should be explicitly converted with properly named methods.

Please then also update the changelog accordingly.

uklotzde (changes requested):

uklotzde:
Reporting a bug and with a follow-up PR that fixes it is nice. Thank you!

mihairadulescu:
Nice!

PR #281, opened 2024-09-11
feat: rename Exception to ExceptionCode

CosminPerRam:
In #279 @uklotzde mentioned that:
> I have noticed that the naming of the types FunctionCode and Exception is inconsistent. Shouldn't Exception better be renamed to ExceptionCode to reflect this duality?

This PR does exactly this, renames `Exception` to `ExceptionCode`.

Also, should associated types be renamed? This is an example:
https://github.com/slowtec/tokio-modbus/blob/a5eb819dd30e21e43c46f0af8c64fbbab2a2e619/src/server/rtu_over_tcp.rs#L224
To become:
```rust
type ExceptionCode = ExceptionCode
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,13 @@
 
 ## v0.15.0 (Unreleased)
 
-- Added `Exception::new`.
+- Added `Exception::Custom` and `Exception::new`.
 
 ### Breaking Changes
 
 - Added `Exception::Custom`.
 - Removed `TryFrom<u8>` and `#[repr(u8)]` for `Exception`.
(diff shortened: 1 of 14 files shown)


uklotzde:
> Also, should associated types be renamed? This is an example:
> 
> https://github.com/slowtec/tokio-modbus/blob/a5eb819dd30e21e43c46f0af8c64fbbab2a2e619/src/server/rtu_over_tcp.rs#L224
> 
> To become:
> 
> ```rust
> type ExceptionCode = ExceptionCode
> ```

I don't think so. The type name `Exception` is appropriate here. It could be any type that could be converted into an `ExceptionCode`.

uklotzde (approved):

PR #285, opened 2024-09-19
tcp client: Use non-atomic type for internal transaction id

uklotzde:


src/service/tcp.rs
```diff
@@ -20,12 +17,18 @@ use crate::{
 
 const INITIAL_TRANSACTION_ID: TransactionId = 0;
 
+fn next_transaction_id(transaction_id: &mut TransactionId) -> TransactionId {
+    let next_transaction_id = *transaction_id;
+    *transaction_id = next_transaction_id.wrapping_add(1);
+    next_transaction_id
+}
+
```
CosminPerRam, line 25: Hey, sorry for hopping in randomly.
If this function is used only in this file, and only in one place (`next_request_hdr`), would it be better to inline it?

uklotzde: The compiler will decide if inlining is appropriate here. No need to clutter the code with `#[inline]`.

https://nnethercote.github.io/perf-book/inlining.html

CosminPerRam: I haven't formed my question well, I meant to do the adding inside that function without making another ope just for that if it has just one usage.

uklotzde: This function is for emphasizing that only a single field is mutated. It is also handy for keeping it close to the constant definition for the initial value.

uklotzde: Management of transaction ids could be decoupled from `Client` in the future.

uklotzde: https://github.com/slowtec/tokio-modbus/pull/285/commits/50bf05e39218cc8bda8fadbda866cc6e4884b4e6

CosminPerRam: Woah, nice stuff, but thats a bit amusing cause it feels like over engineering for just incrementing a number haha

I had a *bit* of fun trying to make a const-only version of it:
```rust
#[derive(Debug)]
struct TransactionIdGenerator {
    next_transaction_id: u16,
}

impl TransactionIdGenerator {
    const fn new() -> Self {
        Self {
            next_transaction_id: 0,
        }
    }
    
    const fn next(self) -> Self {
        Self {
            next_transaction_id: self.next_transaction_id.wrapping_add(1)
        }
    }
    
    const fn now(&self) -> u16 {
        self.next_transaction_id
    }
}

fn main() {
    let mut k = TransactionIdGenerator::new();
    println!("{}", k.now());
    k = k.next();
    println!("{}", k.now());
}
```

uklotzde: > Woah, nice stuff, but thats a bit amusing cause it feels like over engineering for just incrementing a number 

I don't consider modularization and encapsulation as over-engineering. It improves maintainability and reduces the mental load.

In this case the wrapping add is a _strategy_ that only belongs to the generation of transaction ids and is independent from any other aspect of the `Client` struct.

Issue #95, opened 2021-08-20
Add (property-based) fuzzing tests


uklotzde:
...to improve safety and prevent more bugs like #87.

PR #292, opened 2024-10-14
Encode request/response into buffer to avoid temporary allocations

uklotzde:


uklotzde:
The strict validations that are applied now during encoding/decoding might cause issues for devices that don't conform to the spec. Scheduled for v0.16.0 to avoid breaking existing applications.

uklotzde:
Tests for lots of edge cases are missing. These should better be discovered by #95.

src/codec/mod.rs
```diff
@@ -422,60 +460,88 @@ fn coil_to_bool(coil: u16) -> io::Result<bool> {
...
     let mut res = Vec::with_capacity(count.into());
     for i in 0usize..count.into() {
         res.push((bytes[i / 8] >> (i % 8)) & 0b1 > 0);
     }
     res
 }
 
-fn request_byte_count(req: &Request<'_>) -> usize {
+#[cfg(any(feature = "rtu", feature = "tcp"))]
+fn request_pdu_size(req: &Request<'_>) -> io::Result<usize> {
```
flosse, line 490:
```suggestion
fn request_pdu_size(request: &Request<'_>) -> io::Result<usize> {
```

src/codec/mod.rs
```diff
@@ -422,60 +460,88 @@ fn coil_to_bool(coil: u16) -> io::Result<bool> {
...
+            ErrorKind::InvalidInput,
+            "request PDU size exceeded",
+        ));
     }
+    Ok(size)
 }
 
-fn response_byte_count(rsp: &Response) -> usize {
+#[cfg(feature = "server")]
+fn response_pdu_size(rsp: &Response) -> io::Result<usize> {
```
flosse, line 516:
```suggestion
fn response_pdu_size(response: &Response) -> io::Result<usize> {
```

PR #302, opened 2024-11-27
fix: memory leak in tcp client writing holding register

### LINKED ISSUE #301: memory leak in tcp client when repeatedly writing holding register
Hi, I think I found a memory leak bug:

I have a program that makes 2 holding register writes:
first call writes a value
second call writes 2 values (problem happens even with more number of values)

if you keep doing that, program will continue to double the ram needed because of this line in 
src/coded/tcp.rs (line 136) 
      buf.reserve((buf.capacity() - buf_offset) + request_pdu_size + 7);

this happens on every iteration, you will see an initial 8KB usage than 16KB and so on till the program crashes

it does not matter that the server answers, problem will occur eventually

cdbennett:
The call to `BytesMut::reserve()` was passing the total number of bytes we desired as the capacity, but `reserve()` takes only the additional number of bytes we expect to insert. This was resulting in an exponential explosion of the length and a crash after a couple dozen calls at most, with a message like

    memory allocation of 34359738368 bytes failed

Fixes #301


uklotzde (approved):
Thank you for finding and fixing this serious bug!

uklotzde:
RTU was also affected: #304

PR #307, opened 2024-12-12
Disable max. PDU size checks for custom functions

### LINKED ISSUE #306: New MAX_PDU_SIZE limits systems that work beyond that limit via custom functions
v0.16.0 introduced a hardcoded limit on the `MAX_PDU_SIZE`: https://github.com/slowtec/tokio-modbus/compare/v0.15.0...v0.16.0#diff-8dfd85153dbfd0c2fe0b29f7eb8d53fbda43f2f2cc71cfca8fa6402dba78db00R26

But I have a system (labjack, modbus tcp) that has been working with larger packet sizes on v0.15. Their limit is a max ethernet packet size of 1040 bytes. By maintaining this limit, I can reduce the number of sequential reads on the device when doing fancier things like reading N bytes (where N is potentially large) from a `buffer register`: https://support.labjack.com/docs/3-1-modbus-map-t-series-datasheet#id-3.1ModbusMap[T-SeriesDatasheet]-BufferRegisters

They implement a custom modbus function that supports these larger packet sizes: https://support.labjack.com/docs/protocol-details-direct-modbus-tcp#ProtocolDetails[DirectModbusTCP]-ModbusFeedback(MBFB,function#76)

Can the restriction be relaxed, perhaps if using custom modbus functions or when using modbus tcp?

uklotzde:
Fixes #306.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,7 +31,7 @@ repos:
         stages:
           - commit-msg
   - repo: https://github.com/DavidAnson/markdownlint-cli2
-    rev: v0.15.0
+    rev: v0.16.0
     hooks:
       - id: markdownlint-cli2
   - repo: https://github.com/shellcheck-py/shellcheck-py
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 # Changelog
 
+## v0.16.1 (Unreleased)
+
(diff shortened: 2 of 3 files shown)


uklotzde:
The size checks could be omitted for all fixed size requests/responses, because we also check that all bytes have been consumed.

uklotzde:
Confirmed to work as expected. The unconditional check seems to be too restrictive for custom function codes.

PR #314, opened 2025-02-10
Relaxing TCP header verification for devices who don't obey header specs

PoseidonTaylor:


PoseidonTaylor:
Thought I was in forked repository, sorry. Closing PR.
I'll be evaluating whether "Our vendor doesn't obey ModBus spec, but we still need to communicate to their server" is worth escalating to a PR on the main crate.

PR #316, opened 2025-03-14
Remove unnecessary Debug trait requirement

IniterWorker:
The Debug trait was an unnecessary requirement on the abstract transport layer and is unused internally.
Removing it simplifies the code without impacting functionality.

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -3,7 +3,7 @@
 
 //! Modbus clients
 
-use std::{borrow::Cow, fmt::Debug, io};
+use std::{borrow::Cow, io};
 
 use async_trait::async_trait;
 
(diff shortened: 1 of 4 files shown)


IniterWorker:
Just to be thorough, the clippy error are coming from the upstream source code not the modification.

uklotzde:
Some clippy warnings are be caused by the new Rust version and are unrelated. These need to be fixed in a separate PR.

uklotzde:
> Some clippy warnings are be caused by the new Rust version and are unrelated. These need to be fixed in a separate PR.

Should be fixed soon: #317

uklotzde:
I have updated your branch.

uklotzde:
This is a breaking change and requires to be released as v0.17.

What was your issue with the `Debug` bound on this trait? Even though the trait is public it is not intended to be implemented by client applications AFAIR.

IniterWorker:
@uklotzde The issue was mainly a mix of Rust's "orphan rule" and the fact that the Stream abstraction doesn't provide the Debug trait, as seen in serial2-tokio and similar crates. Since tokio-modbus aims to offer an abstract layer for Stream, requiring extensive traits like Debug felt unnecessary when the same functionality can be achieved with fewer constraints.

Hope that provides more clarity to you.

flosse (approved):

PR #334, opened 2025-08-26
Properly Close the tasks for existing connections

wheregmis:
Fixes #333
Maybe #329

Issue #333, opened 2025-08-26
serve_until is not handing current existing clients


wheregmis:
Even when the modbus server is aborted, it responds back with request, so either serve_until is not working, or it has hanging connections

Reproducable Repo:

uklotzde:
Fixed by #336.

Issue #329, opened 2025-08-05
Cancelling or otherwise shutting down the TCP server does not disconnect clients


Simontaga:
I have witnessed an odd issue where even when the TCP server task is cancelled by any means by e.g calling `shutdown` on a `JoinSet` the server is a part of, on-going TCP connections remain open. I can confirm that new connections are not able to be established which is expected.

I _believe_ the culprit is the line linked below, as this lingering task (per connection) appears to keeps parsing frames and passing Modbus function calls down to my `Service`.

https://github.com/slowtec/tokio-modbus/blob/3979a5b2ae1b6a09026c3a41984ac3887989613f/src/server/tcp.rs#L94

I've also used `serve_until` although that does not have any effect and does not close open connections.

PR #336, opened 2025-09-05
Disconnect client sockets when TCP/RTU over TCP server exits (#329)

svc-devel:
I have recently been having problems related to issues #329 and #333 also.  This pull request is a proposed solution that is a slightly more minimal change than the one in #334.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 # Changelog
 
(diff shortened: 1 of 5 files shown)


uklotzde (commented):
Thanks. Looks reasonable.

Please fix the minor formatting issues detected by pre-commit.

Issue #310, opened 2024-12-18
How to specify custom response decoder?


daiguangbao:
it's now trying to get a valid function code (of course there won't be, since the function code is custom), and give up after 20 retries.
tokio-modbus-0.16.1\src\codec\rtu.rs 265 ---> Failed to decode response frame: Invalid function code: 0x65
Is there a way users can specify their own decoder outside?

uklotzde:
Decoding of user-defined function codes (65..=72 and 100..=110) has not been implemented yet. 

`get_response_pdu_len()` only supports a limited subset of functions codes while `decode_response_pdu_bytes()` already passes through all unhandled function codes.

uklotzde:
There is also a considerable amount of redundancy between the code of the mentioned functions. Could probably be rewritten in a better way.

daiguangbao:
yes, when I return a correct number for a custom function code in *get_response_pdu_len()*, it goes correctly.  but there's many other function codes and only the user knowns what len is expected.

uklotzde:
> yes, when I return a correct number for a custom function code in _get_response_pdu_len()_, it goes correctly. but there's many other function codes and only the user knowns what len is expected.

Checking the expected size only is an additional safety measure. The CRC code in the ADU serves as the sentinel.

Decoding of ADU/PDU frames should be rewritten. I have vague ideas, but unfortunately no time to do this.

lindblandro:
@uklotzde Can you expand on your vague ideas about this? I tried mainly changing the default behavior of `get_response_pdu_len()` with
```
-                return Err(Error::new(
-                    ErrorKind::InvalidData,
-                    format!("Invalid function code: 0x{fn_code:0>2X}"),
-                ));
+                if adu_buf.len() >= 3 {
+                    adu_buf.len() - 3
+                } else {
+                    return Err(Error::new(
+                        ErrorKind::InvalidData,
+                        format!("Invalid function code: 0x{fn_code:0>2X}"),
+                    ));
+                }
```
which obviously breaks down if the buffer contains multiple frames.

PR #346, opened 2025-10-01
codec: guess response pdu length for custom fns

lindblandro:
For simple cases (when there's just a single ADU in the buffers) any custom fn PDU frame length can be calculated by subtracting the number of bytes reserved for address and crc (3).

This is an extremely simple solution for the problem, but I've been using this little fix for a while now without issues to implement a custom file transfer function.

#310
#123


lindblandro:
Executed `cargo fmt --all`  to fix the failing pre-commit hook

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 
 ## Unreleased
 
+- Decoding of responses: Guess response length for custom function codes.
```
uklotzde:
```suggestion
- Breaking change (behavior): Guess response length for custom function codes instead of returning an error.
```

uklotzde: Could we be even more specific about how the behavior has changed for users of this crate?

See also: https://github.com/slowtec/tokio-modbus/issues/350#issuecomment-3414353993

uklotzde (changes requested):

CHANGELOG.md
lindblandro: I'd be interested to hear the use cases for relying on frame errors on custom responses in an application, but technically it's true that the behavior has changed. 

Maybe something along the lines of

> Custom function response PDU lengths are calculated by subtracting the mandatory ADU field lengths (unit=1byte, crc=2bytes) from the Modbus ADU frame length.

uklotzde: That's probably too much technical details about how it is actually implemented.

Then let's stick to "... instead of returning an error." as I proposed. This will hopefully be enough to decide if and how the changed behavior affects your use case.

uklotzde (approved):
I am still unsure if this is really has to be considered as a breaking change as proposed in #350. Let's merge it as is until we decide about how to release it.

PR #358, opened 2025-12-11
refactor(server): replace transpose with match for request error hand…

Alarak-Alex:
…ling

Refactor request processing logic in tcp, rtu, and rtu_over_tcp modules to use explicit match expressions instead of transpose. Improve error handling clarity by continuing the loop on error instead of returning immediately.

### DIFF
--- a/src/server/rtu.rs
+++ b/src/server/rtu.rs
@@ -78,12 +78,16 @@ where
     S::Request: From<RequestAdu<'static>> + Send,
 {
     loop {
-        let Some(request_adu) = framed.next().await.transpose().inspect_err(|err| {
-            log::debug!("Failed to receive and decode request ADU: {err}");
-        })?
-        else {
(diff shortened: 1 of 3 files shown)


src/server/rtu.rs
```diff
@@ -78,12 +78,16 @@ where
...
-            log::debug!("Stream has finished");
-            break;
+        let request_adu = match framed.next().await {
+            Some(Ok(adu)) => adu,
+            Some(Err(err)) => {
+                log::debug!("Failed to receive and decode request ADU: {err}");
+                continue;
+            }
+            None => {
+                log::debug!("TCP socket has been closed");
```
sahil-boom: I think this log messaged should be `"Stream has finished"`

flosse:
I'm closing this issue here. Feel free to re-open it if necessary.

Issue #324, opened 2025-06-11
Timeout causes transaction ID mismatch with reused context


thalesmaoa:
### Environment
- 10x RTU devices → RS485 → USR-RD302 TCP converter → tokio-modbus (TCP)
- Library version: tokio-modbus (latest)
- OS: (e.g., Linux)

### Problem
When a Modbus request times out due to device unavailability, subsequent requests using the same context fail with `mismatching headers` errors when delayed responses arrive. 

**Error example:**
```bash
Err: mismatching headers: expected/request = Header { transaction_id: 97, unit_id: 2 }, actual/response = Header { transaction_id: 95, unit_id: 2 } Ok(ReadHoldingRegisters([17166, 45875, 17162, 45875, 17158, 13107, 17267, 45875, 17260, 19661, 17263, 52429, 16925, 39322, 16965, 28836, 16840, 62915, 16373, 49807, 49300, 48234, 49344, 41943, 49131, 34079, 49480, 8389, 16451, 38273, 49060, 23069, 48881, 43516, 16294, 26214, 16563, 63439, 16603, 8913, 16471, 36176, 16765, 28836, 48979, 29884, 48993, 1573, 48907, 50856, 48969, 64487, 17008, 7864, 17357, 15729, 17506, 34079, 17367, 7864, 17303, 41943, 17632, 15401, 17572, 37356, 50167, 52429, 17463, 24904, 17149, 60293, 17216, 47186, 17261, 28836]))
```

### Reproduction Steps
1. Shared context used for multiple devices
2. Device responds with high latency
3. Timeout triggers correctly
4. New request uses same context
5. Old delayed response arrives → header mismatch

### Current Implementation

```rust
let mut locked_ctx = ctx.lock().await;  // pegando lock mutável assíncrono
locked_ctx.set_slave(tokio_modbus::Slave(self.id));
let reg = tokio::time::timeout(
    timeout,
    async {
        let result = locked_ctx.read_holding_registers(6, 76).await;
        result
    }
).await;
```

Need a way to discard pending responses after timeout when reusing contexts. Suggested solutions:

    1. Context reset method: ctx.reset() (clears buffers/transaction state)
    2. Automatic response discard on timeout
    3. Transaction ID invalidation API

thalesmaoanz:
Does the only solution is to flush tcp stream? If is that so, I need to disconnect and reconnect.

I've made a huge complicated workaround.

```rust
let reg = tokio::time::timeout(
    timeout,
[...]

PR #359, opened 2026-01-11
Discard invalid TIDs and retry read

rlarsson-mc:
Since the stream is owned by the service and no other thread can call while sending a request, it's safe to discard invalid TIDs as there won't be any pending requests that could be affected.

Fix for #324

### DIFF
--- a/src/service/rtu.rs
+++ b/src/service/rtu.rs
@@ -65,20 +65,38 @@ where
         framed.read_buffer_mut().clear();
         framed.send(req_adu).await?;
 
-        let res_adu = framed
-            .next()
-            .await
-            .unwrap_or_else(|| Err(io::Error::from(io::ErrorKind::BrokenPipe)))?;
-        let ResponseAdu {
-            hdr: res_hdr,
-            pdu: res_pdu,
-        } = res_adu;
-        let ResponsePdu(result) = res_pdu;
+        let mut mismatch_err: Option<ProtocolError> = None;
+        let result = loop {
+            let res_adu = match framed.next().await {
+                Some(Ok(adu)) => adu,
+                Some(Err(e)) => return Err(e.into()),
+                None => match mismatch_err {
+                    Some(mismatch_err) => return Err(mismatch_err.into()),
+                    None => {
+                        return Err(io::Error::new(io::ErrorKind::BrokenPipe, "broken pipe").into())
+                    }
+                },
+            };
 
-        // Match headers of request and response.
-        if let Err(message) = verify_response_header(&req_hdr, &res_hdr) {
-            return Err(ProtocolError::HeaderMismatch { message, result }.into());
-        }
+            let ResponseAdu {
+                hdr: res_hdr,
+                pdu: res_pdu,
+            } = res_adu;
+            let ResponsePdu(response) = res_pdu;
+
+            // Match headers of request and response.
+            if let Err(message) = verify_response_header(&req_hdr, &res_hdr) {
+                log::debug!(
+                    "Header mismatch expected {message}, detected {response:?}, retrying..."
+                );
+                mismatch_err = Some(ProtocolError::HeaderMismatch {
+                    message,
+                    result: response,
+                });
+            } else {
+                break response;
+            }
+        };
 
         // Match function codes of request and response.
         let rsp_function_code = match &result {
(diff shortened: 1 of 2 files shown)


flosse (changes requested):
Please avoid unwraps and unnecessary mutable values.

CosminPerRam:
Bump @flosse.
I've also encountered this issue, any thoughts on merging this?
I'd suggest to make this behavior configurable (i.e. `.set_ignore_mismatches(true)`) as one could still want tight checks regarding the stream flow.

rlarsson-mc:
@CosminPerRam 
> Bump @flosse. I've also encountered this issue, any thoughts on merging this? I'd suggest to make this behavior configurable (i.e. `.set_ignore_mismatches(true)`) as one could still want tight checks regarding the stream flow.

I would say it should if so be the other way around, as this should be implemented as default, as I interpret the [standard atleast](https://www.modbus.org/file/secure/messagingimplementationguide.pdf):
> When a response frame is received on a TCP connection, the Transaction Identifier
carried in the MBAP header is used to associate the response with the original request
previously sent on that TCP connection:
 If the Transaction Identifier doesn't refer to any MODBUS pending transaction, the
response must be discarded.
 If the Transaction Identifier refers to a MODBUS pending transaction, the response
must be parsed in order to send a MODBUS Confirmation to the User Application
(positive or negative confirmation) 

I interpret that as that by default a faulty TID should not interfere with the function of the client.
I also don't think `tokio-modbus` can handle sending more than one message at a time for TCP, as the client itself holds the stream and it's held during the whole duration of a transaction, which is how it needs to be for RTU, and can't share the stream with other TCP instances (but correct me if I'm wrong).
If so, there shouldn't be more than 1 active TID at a time, it can safely be discarded without having to buffer it for any other request.

That said, it could be specific devices that should fail on faulty TIDs, so I would if so suggest the inverse instead a `.set_fail_on_mismatch` or something. Or just have `set_ignore_mismatch` and as default true.

Tuetuopay:
> I interpret that as that by default a faulty TID should not interfere with the function of the client.

Correct.

> I also don't think `tokio-modbus` can handle sending more than one message at a time for TCP, as the client itself holds the stream and it's held during the whole duration of a transaction, which is how it needs to be for RTU, and can't share the stream with other TCP instances (but correct me if I'm wrong). If so, there shouldn't be more than 1 active TID at a time, it can safely be discarded without having to buffer it for any other request.

Correct with the current implementation.

However. I just rewrote the stream handling part of the library which unlocks concurrent requests to the stream, acheived with the underlying connection being handled by a dedicated Tokio task. Interaction with the connection goes through channels, replies are sent back to awaiting clients with oneshot channels. It's a bit similar to what's done in rumqtt. This gains a few things:

- The task keeps track of the transaction IDs, to notify the correct waiter on the connection. This completely eliminates #324 and uses transaction IDs for their actual purpose.
- The clients properly handles out-of-order replies from a server.
- The client can be cheaply cloned and multiple concurrent tasks can use it to send concurrent requests.

With this system, we could even have the task manage the full lifecycle of the connection, automatically reconnecting when the connection breaks (this would however require an even bigger refactor, where the library always get handled a connector function instead of possibly an established stream).

To keep things simple and the API consistent, the RTU client got also changed to wrap the stream in a mutex. To library users, nothing really changes aside from the `&mut` requirement on the client, and it being cheaply clonable. This, however, gains one feature for both TCP and RTU: multiple slave IDs can be queried without changing the ID each time, as it's a property of the handle.

In the same vein, the TCP server also got a rewrite to unlock concurrent requests. In addition to spawning a Tokio task for each client, it also spawns a task for each request (like every web framework like Axum would do). When writing Modbus/TCP servers that are gateways (i.e. forward calls to other modbus devices), a client connection won't be blocked if one of the backend device is slow to answer.

The branch can be found [here](https://github.com/Tuetuopay/tokio-modbus/tree/concurrent-requests).

Since it's a pretty big change, @flosse and other maintainers, how would you like to proceed? I'd like to contribute back, but I don't want to overwhelm you with a PR that big without discussing this further. And, also, if the API changes are fine with you.

flosse:
@rlarsson-mc and @Tuetuopay, unfortunately I don't have the capacity right now to review this in detail (see #380), but I also don't want to hold up the process. So here's a quick question: What helps you make a decision or move forward here?

Issue #325, opened 2025-06-18
Error: "The operation completed successfully." In tcp client example.


paul-fornage:
When running the example code, I am getting this when the slave device does not respond in time (on the order of half a second)

```
thread 'main' panicked at src\main.rs:12:45:
called `Result::unwrap()` on an `Err` value: Transport(Os { code: 0, kind: Uncategorized, message: "The operation completed successfully." })
```

I narrowed down the source of the error object to line below.
https://github.com/slowtec/tokio-modbus/blob/56bf0fa650a6d7d986030da72a15ea751b32fa30/src/service/tcp.rs#L104

It should be said that I only had it happen on Windows, but on my main Linux machine not only did I not get that unhelpful message, but the error wouldn't happen at all. Not too surprising considering the message comes from a syscall. I'm less concerned about the fact that an error happened than I am about the unhelpful message. If you want me to try something to test a fix I will gladly do it, but I would suggest not relying on the 'get last error' syscall to trace the error. Or maybe wrap it with some extra context like "__Failed to read response frame with OS error: {}__"

PR #360, opened 2026-01-16
Use explicit error message instead of syscall for unexpected TCP connection drops

sahil-boom:
The use of `io::Error::last_os_error` [can result is confusing error messages](https://doc.rust-lang.org/std/io/struct.Error.html#method.last_os_error) such as an indication that operation completed successfully. So, I've replaced the `io::Error::last_os_error` call with an explicit `UnexpectedEof` error. I believe the only case where we should get an error from the `next()` call is if we have a premature connection closure.

This is a fix for https://github.com/slowtec/tokio-modbus/issues/325.

### DIFF
--- a/src/service/tcp.rs
+++ b/src/service/tcp.rs
@@ -100,7 +100,12 @@ where
         framed.read_buffer_mut().clear();
         framed.send(req_adu).await?;
 
-        let res_adu = framed.next().await.ok_or_else(io::Error::last_os_error)??;
+        let res_adu = framed
+            .next()
+            .await
+            .ok_or_else(|| {
+                io::Error::new(io::ErrorKind::UnexpectedEof, "connection closed by remote")
+            })??;
         let ResponseAdu {
             hdr: res_hdr,
             pdu: res_pdu,


sahil-boom:
Hi @flosse I'm sure you're super busy, but I wanted to bump this PR since it's a pretty straightforward fix. One of my colleagues ended up in an extended troubleshooting session with our controller because of the misleading error messages sending him down the wrong path.

flosse:
> Hi @flosse I'm sure you're super busy

Yes, that's true. And unfortunately, there's currently no one supporting our open-source work like this (nor is there any project focused on Modbus).
My reviews are therefore minimal (if not nonexistent).
So I'm leaving the responsibility to the community.

sahil-boom:
> > Hi @flosse I'm sure you're super busy
> 
> 
> 
> Yes, that's true. And unfortunately, there's currently no one supporting our open-source work like this (nor is there any project focused on Modbus).
> 
> My reviews are therefore minimal (if not nonexistent).
> 
> So I'm leaving the responsibility to the community.
> 
> 

Fair enough! I appreciate all your hard work in getting this project to the state it is.

flosse:
BTW: #380

PR #361, opened 2026-02-18
fix(rtu): replace decode loop with rolling buffer scan

obayemi:
Current implemention of RTU server has some issues in real world situations:
- if an other actor on the bus writes slowly, sending the messages fragmented, the current implementation will drop the start of the payload before receiving the end even for valide potential frames.
- if some data e.g. a big write frame or multiple read response is sent on the bus, if some of the bytes did accidentally look like a modbus frame, with invalid CRC, the decoder would skip the whole frame even if the start of the actual frame was in the bytes contained in this erroneous detection.

this made the implementation unusable for buses with multiple other slaves.

this PR does:
Never mutate buffer during scanning — use read-only slices to check all byte offsets for valid CRC-confirmed frames. Always return Ok(None) instead of Err to keep FramedRead stream alive with partial data.

Add tests for fragmented arrival, back-to-back frames, garbage with valid function codes but bad CRC, partial frame preservation, all-garbage buffers, no Err on excessive garbage, and deep frame recovery.

### DIFF
--- a/src/codec/rtu.rs
+++ b/src/codec/rtu.rs
@@ -37,38 +37,21 @@ impl Default for FrameDecoder {
 }
 
 impl FrameDecoder {
-    pub(crate) fn decode(
-        &mut self,
-        buf: &mut BytesMut,
-        pdu_len: usize,
-    ) -> io::Result<Option<(SlaveId, Bytes)>> {
-        const CRC_BYTE_COUNT: usize = 2;
-
-        let adu_len = 1 + pdu_len;
-
-        if buf.len() < adu_len + CRC_BYTE_COUNT {
-            // Incomplete frame
-            return Ok(None);
+    fn track_skipped(&mut self, bytes: &[u8]) {
+        for &b in bytes {
+            if self.dropped_bytes.len() >= MAX_FRAME_LEN {
+                log::error!(
+                    "Dropped {} byte(s) without finding a valid frame: {:X?}",
+                    self.dropped_bytes.len(),
+                    self.dropped_bytes
+                );
+                self.dropped_bytes.clear();
+            }
+            self.dropped_bytes.push(b);
         }
+    }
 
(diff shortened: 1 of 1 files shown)


src/codec/rtu.rs
```diff
@@ -839,5 +841,146 @@ mod tests {
...
+            let res = decoder.decode(&mut buf).unwrap();
+            assert!(res.is_none());
+            // Most garbage discarded; last byte may remain as potential frame start
+            assert!(buf.len() <= 1);
+        }
+
+        #[test]
+        fn decode_never_returns_err_with_excessive_garbage() {
+            // The old codec returned Err after 20 retries, killing FramedRead.
+            // The new codec must always return Ok(None).
```
flosse: Talking about "old" and "new" doesn't make sense here.

src/codec/rtu.rs
```diff
@@ -280,42 +239,85 @@ impl Decoder for ResponseDecoder {
...
+                // Invalid function code at this offset — confirmed garbage
+                continue;
+            }
+            Ok(Some(len)) => len,
+        };
 
-        if let Err(err) = result {
-            log::warn!("Failed to decode {pdu_type} frame: {err}");
-            frame_decoder.recover_on_error(buf);
+        let adu_len = 1 + pdu_len;
```
flosse: For better readability the `1` could be a named constant.

obayemi:
Hey ! any news ? is there a planned timeline for the 0.18.0 milestone ?

flosse:
> Hey ! any news ? 

Did you see my comments?

> is there a planned timeline for the 0.18.0 milestone ?

see #380

obayemi:
> > Hey ! any news ?
> 
> Did you see my comments?
> 

I didn't see any comment other than your initial one ... 



> > is there a planned timeline for the 0.18.0 milestone ?
> 
> see #380

:+1:

PR #368, opened 2026-03-11
Implement RTU Broadcast

fpagliughi:
This implements the broadcast feature for RTU. By the spec, if the Slave ID = 0, then all devices should process and execute the request, but none should put a response on the bus.

So, this PR updates the RTU clients and servers:

- For the client: If the request is a broadcast, the client sends the request, but does not await a response.
- For the server: When the request is received, it is processed, but no response is sent.

I am fairly certain this does not pertain to TCP clients and servers, so the behavior there was not changed. The specs say that zero is valid and normal for a TCP transaction.

I was not sure what to do about RTU over TCP. I implemented the broadcast behavior. But if that is not correct, I can remove it.

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -22,7 +22,10 @@ pub mod sync;
 #[async_trait]
 pub trait Client: SlaveContext + Send {
     /// Invokes a _Modbus_ function.
-    async fn call(&mut self, request: Request<'_>) -> Result<Response>;
+    ///
+    /// For broadcast requests (slave ID 0), the request is sent but
+    /// no response is awaited, and `Ok(Ok(None))` is returned.
+    async fn call(&mut self, request: Request<'_>) -> Result<Option<Response>>;
 
     /// Disconnects the client.
     ///
@@ -130,7 +133,7 @@ impl From<Context> for Box<dyn Client> {
 
 #[async_trait]
 impl Client for Context {
-    async fn call(&mut self, request: Request<'_>) -> Result<Response> {
+    async fn call(&mut self, request: Request<'_>) -> Result<Option<Response>> {
         self.client.call(request).await
     }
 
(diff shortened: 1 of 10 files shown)


src/client/mod.rs
```diff
@@ -22,7 +22,10 @@ pub mod sync;
 #[async_trait]
 pub trait Client: SlaveContext + Send {
     /// Invokes a _Modbus_ function.
-    async fn call(&mut self, request: Request<'_>) -> Result<Response>;
+    ///
+    /// For broadcast requests (slave ID 0), the request is sent but
+    /// no response is awaited, and `Ok(Ok(None))` is returned.
+    async fn call(&mut self, request: Request<'_>) -> Result<Option<Response>>;
```
flosse, line 28: That's a pretty drastic change for almost all calls, just to cover that one special broadcast case.
Why not a separate function?
```rust
async fn broadcast(&mut self, request: Request<'_>) -> Result<()>;
```

fpagliughi: Yeah, actually, adding a `broadcast()`  was the first thing I thought of doing. But after considering it for a few minutes, the separate broadcast function seemed like a hack. I really didn't like the idea of leaving the normal _call()_ pathway open to processing IDs of zero incorrectly. At best, it seems like it might confuse users; at worst it seems like a bug.

So I asked a few of the people who would be using this in a system that regularly does broadcasting, and they all stated that they would prefer to just use the library "as-is" with a zero ID. So, for them, the matter of a bigger change was different from looking at the library from the inside or the outside.

Finally, for me, as a _user_, I agree with them. I'm writing a network MQTT bridge to Modbus RTU. I would need to intercept every incoming request for the bus, check the ID, and take a separate path for ID==0. It seems like a smaller change for the library is a bigger change for users.

flosse: good point!

PR #371, opened 2026-04-07
[fix]add socket address parameter for the error callback function of tcp/r…

wngxao:


flosse:
@wngxao this looks good to me but would you mind to fix the issues in the pre-commit?

wngxao:
@flosse OK

Issue #374, opened 2026-04-15
Having \0 first character in rtu buffer leads to "Too many retries" error


KodyVB:
I am spawning a `tokio_modbus::server::rtu` server and using its via:

```rs
let _server = tokio::spawn(async move {
    server_serial.clear(tokio_serial::ClearBuffer::All).unwrap();
    let server = Server::new(server_serial);
    let service = Service {
        modbus_tables: shared_modbus,
        board_type: board_info.board_type,
    };
    server
        .serve_forever(service)
        .await
        .unwrap_or_else(|e| match e {
            _ => {
                eprintln!("{e:?}");
            }
        });
});
_server.await.unwrap();

```
(I'm trying to avoid posting unneeded code, but if it would help to post more, please let me know)

I found that sometimes, this would seem to get stuck. I'd go to the `call` function I was defining in the `Service` (following the examples from this repo) and try to print everything that was getting called, but it would get stuck before the calls. So, I dug through the source code until I found by printing `buf` inside `decode`, if the buffer started with `\0`: it would grow until I got a "Too many retries" error.

A log of the behavior is provided here (the RHR lines are in the Service call function to print Read Holding Register calls): [output.log](https://github.com/user-attachments/files/26764278/output.log)

PR #375, opened 2026-04-15
fix(rtu) Remove any leading 0's from buffer during decode

KodyVB:
This was intended to solve [this issue](https://github.com/slowtec/tokio-modbus/issues/374)

It seems to have fixed the issues I was having, but I don't know if it would be good as a general-purpose solution. I'm definitely not an expert when it comes to this, but I figured my solution might be able to help narrow things down even if it's wrong.

### DIFF
--- a/src/codec/rtu.rs
+++ b/src/codec/rtu.rs
@@ -300,6 +300,10 @@ where
 {
     const MAX_RETRIES: usize = 20;
 
+    while buf.starts_with(&[b'\0']) {
+        drop(buf.split_to(1));
+    }
+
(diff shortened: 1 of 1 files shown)


KodyVB:
After looking more into it, Modbus protocol shows that the beginning of a message is marked by a silent interval of at least 3.5 character times. Slave addresses (the first part sent in a message) are usually non-zero numbers, but some use a slave address of 0 as a broadcast message. So, this solution worked for me since I'm not expecting broadcast messages, but it probably won't be a good general-purpose solution.

PR #376, opened 2026-04-28
add socket address parameter for the error callback function of tcp/rtu-over-tcp server

wngxao:


flosse:
@wngxao! Thanks for your suggested change.

Could you add a short PR description explaining the motivation?
And a problem I see is, that it breaks the existing `on_process_error` signature for all callers — could we avoid that?

Issue #377, opened 2026-05-25
RTU server does not support slave ID filtering at the transport level


ostrufka:
I have some applications that act as Modbus RTU slaves on shared RS-485 buses. Multiple devices coexist on the same serial line, each with a unique slave ID. The servers must respond only to their assigned slave IDs and stay completely silent for all others.

I have integrated tokio-modbus into my applications, but I realized that the current implementation requires that all those slave IDs filtering be done in the application layer. So, for every appication I have to include a piece of code to deal with this. In my opinion and according to the Modbus RTU specification, slave ID filtering is a transport-level responsibility.  

I evaluated both older and current versions of the create and found some changes regarding this topic.

  - v0.11.0: SlaveRequest exposed the slave ID to the application, but the Service trait returned Result<Response, Exception> — there was no way to return "no response". Every request forced a response, so the server could never stay silent for requests meant for other devices.
  - v0.16.1+ (and current v0.17.0): The Service trait was improved with type Response: Into<Option<Response>>, allowing applications to return Ok(None) to suppress a response. However, slave ID filtering is pushed entirely to the application layer — the Service implementation must check the slave ID and decide whether to respond. This works, but places a transport-level concern in the wrong layer.

While researching, I also found several related older issues that show this has been a recurring gap:

  - #96 — First report: "Server::Service trait has no mechanism to obtain request's unit ID" (2021)
  - #141 — Added SlaveRequest as an application-layer workaround
  - #264 — The Ok(None) support was accidentally broken during a refactoring because no tests covered this use case. As noted by the maintainer: "Looks like this feature got lost during the refactoring. It was informally proposed as a solution, but no one has provided a test for this use case."
  - #315 — User struggling with multiple slaves on the same RS-485 bus

Based on that, I have implemented a potential solution for this issue and I would like to know if this approach aligns with the project's direction and if you'd be open to a PR.

In summary, I added a with_slave_id() builder method to the RTU Server that filters requests before they reach the application:                           
                                                                                                                                               
  // Server only responds to requests addressed to slave 1.
  // Requests for any other slave ID are silently ignored.
  let server = Server::new(serial).with_slave_id(Slave(1));
  server.serve_forever(my_service).await?;

  // Without with_slave_id(), the server responds to all requests (current behavior, unchanged).
  let server = Server::new(serial);
  server.serve_forever(my_service).await?;

[...]

PR #378, opened 2026-05-29
Add transport-level slave ID filtering to RTU server

ostrufka:
# Summary

  - Adds a with_slave_id() builder method to the RTU Server that filters requests at the transport level before they reach the Service
  - Requests for non-matching slave IDs are silently ignored, as required by the Modbus RTU specification
  - Fully backwards compatible — without with_slave_id(), the server accepts all requests (current behavior unchanged)

  Closes Issue #377

  # Changes

  src/server/rtu.rs:
  - Added slave_id: Option<SlaveId> field to Server
  - Added with_slave_id() builder method
  - Added filtering logic in process(): requests with mismatched slave ID are skipped via continue
  - Made process() generic over T: AsyncRead + AsyncWrite + Unpin (consistent with rtu_over_tcp::process) to enable unit testing without serial hardware

  examples/rtu-server-address.rs:
  - Simplified to use with_slave_id() instead of application-layer filtering in Service::call

  # Tests

  3 new unit tests:
  1. responds_to_matching_slave_id — server responds normally
  2. ignores_mismatched_slave_id — server stays silent (timeout)
  3. without_slave_id_responds_to_all — backwards compatibility confirmed

  All existing CI checks pass (fmt, clippy, doc, 118 tests total).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@
 
 ## Unreleased
 
+- Server: Add transport-level slave ID filtering to the RTU server via the
+  `Server::with_slave_id()` builder method. Requests addressed to other slave
+  IDs are silently ignored, as required by the Modbus RTU specification
+  ([#378](https://github.com/slowtec/tokio-modbus/pull/378)).
+
 ## v0.17.0 (2025-10-22)
 
 - Breaking change (behavior): Guess response length for custom function codes
--- a/examples/rtu-server-address.rs
+++ b/examples/rtu-server-address.rs
@@ -1,42 +1,34 @@
 // SPDX-FileCopyrightText: Copyright (c) 2017-2026 slowtec GmbH <post@slowtec.de>
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
-//! RTU server example with slave address filtering and optional response
+//! RTU server example with transport-level slave address filtering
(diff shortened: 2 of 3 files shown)


ostrufka:
I noticed that one of the CI checks have failed (taplo-lint).

This pre-commit failure isn't related to the PR itself. The `taplo-lint` hook is failing to download the schema catalog from schemastore.org, which changed its format recently and breaks all current taplo releases (see tamasfe/taplo#799 and ComPWA/taplo-pre-commit#35). 

Every PR here since the end of April will fail the same way, and re-running won't help until a fixed taplo is released.
All the other hooks (fmt, clippy, cargo-doc, etc.) pass.

A possible workaround until then is overriding the hook args in `.pre-commit-config.yaml`:

      - id: taplo-lint
        args: [] # don't fetch the schemastore.org catalog

flosse:
@ostrufka Thanks for your PR! Sorry for the late response (see #380).
Would you mind to add a note to the CHANGELOG.md?
thx!

flosse (approved):
As far as I can see, this looks fine ☺️
Thank you!

ostrufka:
Thanks @flosse for the reply. 

I just added a CHANGELOG entry under ## Unreleased. 

Thanks for the review!

